Issue #1, opened 2021-08-16
Add bars to show door position


zbyju:
2 bars that move up and down depending on the position of the doors

PR #44, opened 2026-08-31
feat: add ubuntu provisioning script install-all.sh

zbyju:
## What

New `source/apps/startup/scripts/ubuntu/install-all.sh` — a rewritten, idempotent provisioning script for fresh Ubuntu (22.04+) panel computers. The old `installAll.sh` is kept for now and can be removed once the new script is proven on a real machine.

## What it does

1. Preflight: refuses root, checks Ubuntu >= 22.04, x86_64, and internet + DNS (clear Czech error if the network is missing), gets sudo once and keeps it alive
2. `apt update` + non-interactive `upgrade`
3. Base utilities: curl, git, jq, unzip + network debugging tools (net-tools, dnsutils, traceroute, mtr, nmap, ethtool) + htop/vim/tree/lsof + fonts-open-sans
4. TeamViewer (amd64 deb) + disables Wayland in gdm3
5. Node.js 18.12.1 via `n`, npm global prefix in `~/.npm-global` (PATH lines added only once)
6. Removes update-manager
7. Wine — now non-interactive (`-y`), fixing the hang in the old script
8. Clones/updates `babybox` + `BB`; the GitHub token is auto-read from `github-token.txt` on a flash drive if present, otherwise git prompts (Czech message explains Bitwarden + flash drive)
9. Firefox: snap install/refresh + `/etc/firefox/policies/policies.json` with force-installed [AutoFullscreen](https://addons.mozilla.org/en-US/firefox/addon/autofullscreen/) (`{d320c473-63c2-47ab-87f8-693b1badb5e3}`) and autoplay allowed for all sites (`Permissions.Autoplay.Default = allow-audio-video`, locked)
10. Autostart `.desktop` entry (paths from `$HOME`, no hardcoded `/home/babybox`)
11. Cron watchdog (same `internet_check.sh` logic, idempotent crontab entry)
12. Disables screen blanking/lock/suspend via gsettings
13. pnpm 7.5.0 + `pnpm install` + `node src/index.js --install --ubuntu` (inlined from `install.sh`, without its trailing `$SHELL` that would hang a piped run)

All user-facing output is in Czech, with numbered steps and explicit "action required" callouts (sudo password, GitHub token).

## Verified

- `bash -n` and shellcheck clean (info level included)
- Generated `policies.json` validated with jq; extension ID extracted from the real AMO `.xpi`
- Snap Firefox reads `/etc/firefox/policies/policies.json` (Firefox >= 91)

**Not yet tested on a real Ubuntu machine** — needs one full run + a second run (idempotency) on the first machine before rolling out to the fleet.


zbyju:
## Static review against the provisioning requirements

**Verdict: 24 of 25 requirements pass. One partial (req 9 — non-interactive), caused by the single real bug found: `DEBIAN_FRONTEND=noninteractive` never reaches apt, because sudo strips it.** Everything else is minor. Reviewed statically — not run on a machine. Line numbers refer to `install-all.sh`.

### The one real bug (can hang a first run on some hardware)

**`export DEBIAN_FRONTEND=noninteractive` / `NEEDRESTART_MODE=a` (L32–33) do not survive `sudo`.** Ubuntu's default sudoers has `env_reset`, and neither variable is in `env_keep`. Every `sudo apt-get` call (L159–160, 184, 201, 226, 264, 278) runs with the default interactive debconf frontend.

- The `upgrade` is protected against *conffile* prompts by `--force-confdef/confold` (L160–162), but not against *debconf* questions. Concrete failure: a BIOS-boot machine getting a grub-pc "install to which device?" dialog during `apt-get upgrade` — the run hangs, and when piped into bash there is no usable stdin to answer it.
- The other installs (base packages, TeamViewer, npm, wine) have neither the env var nor the force-conf options.
- Fix: a small wrapper, e.g. `apt_install() { sudo env DEBIAN_FRONTEND=noninteractive NEEDRESTART_MODE=a apt-get ...; }`, used everywhere. Falsifiable by `sudo env | grep DEBIAN` on any stock Ubuntu box.

Note: the *old* script's wine hang was simply the missing `-y` (`sudo apt install wine`); the new `-y -q` (L278) fixes that specific case even without the env fix.

### Requirements

| # | Requirement | Verdict | Evidence |
|---|---|---|---|
| 1 | Root/OS/arch checks | Pass | L100–125: refuses root, `ID=ubuntu`, major ≥ 22, x86_64, clear Czech messages |
| 2 | No destructive ops | Pass | None present. Note: `chown -R $USER /usr/local/{bin,lib,include,share}` (L252–253) is invasive but inherited from the old script and not destructive |
| 3 | Hardware-agnostic | Pass | Only the arch check |
| 4 | No hardcoded user/home | Pass | `$HOME`/`$USER` throughout, incl. `.desktop` (L397) and cron (L408). Minor: `$USER` + `set -u` dies ungracefully if unset; `$(id -un)` is safer |
| 5 | No secrets | Pass | Token read from flash drive or prompted; only an email constant (L23) |
| 6 | Fail clearly | Pass | L19 `set -Eeuo pipefail`; ERR trap prints failed step name + line + what to do (L69–80) |
| 7 | Net/DNS pre-check | Pass | L127–141: `getent` DNS + wget spider on 4 hosts. All 4 return 200 after redirects (verified), so no false failures. Gap: nodejs.org, npm registry, and archive.ubuntu.com aren't pre-checked — a partial firewall passes the check and fails mid-run (the trap handles it, but the up-front explanation only covers the 4 hosts) |
| 8 | Idempotent | Pass | PATH grep-guarded (L236), cron rebuilt via `grep -v` (L410), clone-or-pull (L322), configs overwritten, package/version checks everywhere. Only soft spot: `git pull` fails on a diverged/dirty clone — clear error, not breakage |
| 9 | Only sudo + GitHub prompts | **Partial** | Both prompts announced (L143, L313–319). But the sudo env bug above means apt/debconf can still go interactive. Tiny edge: the GitHub announcement is skipped if `~/.git-credentials` exists without a github.com entry (L312) |
| 10 | Pipe-safe | Pass | No `read`, no `$SHELL` (the old `install.sh` trailing `$SHELL` was deliberately dropped — inlined at L434–442); sudo and git prompt on `/dev/tty`, not stdin |
| 11 | Installs own deps | Pass | Uses only wget/getent/tput up front; curl/git/jq installed (L170–175) before first use (L285, L381) |
| 12 | Czech output | Pass | All output Czech (ASCII, no diacritics — acceptable) |
| 13 | Numbered steps + callouts | Pass | `[n/13]` step() (L44–49), `action_required` banners (L55–60); TOTAL_STEPS=13 matches 13 calls |
| 14 | Update + base/net tools | Pass | net-tools (ifconfig), dnsutils, traceroute, mtr, nmap, ethtool, curl (L170–175) |
| 15 | TeamViewer + Wayland off | Pass | L192–218. Minor: if `custom.conf` had no `[daemon]` section, the fallback `sed` inserts nothing yet the script still prints "Wayland vypnut" (L210–214) — re-check after the fallback |
| 16 | Node 18.12.1, pnpm 7.5.0, sudo-free globals | Pass | `n` + `~/.npm-global` prefix persisted via npmrc (L230–241); pnpm pinned (L434–437); panel globals (pm2 etc.) installed by `node src/index.js --install` through that prefix |
| 17 | update-manager removed | Pass | L262–269 |
| 18 | Wine non-interactive | Pass | `-y -q` (L278); fully robust only after the env fix |
| 19 | Repo clones + BB auth | Pass | Bitwarden + account + flash-drive file all named in the message (L313–319); auto-pickup from `/media/$USER/*` and `/media/*` with nullglob (L289–297); correct Ubuntu mount path |
| 20 | Firefox snap | Pass | Install or refresh, refresh failure downgraded to a warning (L347–356) — right call, since refresh fails when Firefox is running |
| 21 | Enterprise policies | Pass | `/etc/firefox/policies/policies.json` is the documented path the snap reads (system-files interface); extension ID `{d320c473-63c2-47ab-87f8-693b1badb5e3}` matches the live AMO xpi manifest (verified); `Permissions.Autoplay.Default=allow-audio-video, Locked` is the correct all-sites autoplay policy; JSON jq-validated before install (L381) |
| 22 | `.desktop` autostart, no systemd | Pass | L391–401, keeps startup.sh/pm2 |
| 23 | Cron watchdog | Pass | Same `internet_check.sh` (weekly Monday-12:00 reboot + gateway-ping reboot), root crontab rebuilt idempotently (L405–413); `grep -v` also cleans the old hardcoded `/home/babybox` entry — good migration detail |
| 24 | No blank/lock/suspend | Pass | L418–428. Note: on failure it only warns, yet the final summary (L461) still claims the screen won't blank — make the summary honest or fail the step |
| 25 | Firefox launch untouched | Pass | Script changes config only; launch stays in the unchanged startup/BB code |

### Smaller notes (no requirement broken)

- `mktemp -d` dir is leaked; only the deb inside is deleted (L196, L202).
- `chmod 666` on `/var/restart_lock` and the log (L407) — world-writable root-owned files; harmless on a single-user kiosk, but `644` would do since cron runs as root anyway.
- First Firefox launch needs network to fetch the force-installed extension — fine, since the watchdog reboots on lost connectivity anyway, but worth checking `about:policies` on machine #1, as the PR already plans.

The PR's own caveat stands: nothing replaces one full run + one re-run on the first real machine, especially step 13 (the panel's own installer) and the policies pickup.

References: [Ubuntu: customize the Firefox snap with enterprise policies](https://discourse.ubuntu.com/t/how-to-customize-the-firefox-snap-with-enterprise-policies/23074), [Bugzilla 1717216](https://bugzilla.mozilla.org/show_bug.cgi?id=1717216), [Mozilla: Customizing Firefox using policies.json](https://support.mozilla.org/en-US/kb/customizing-firefox-using-policiesjson)

zbyju:
Addressed in e30acab:

- **sudo env bug (the req 9 partial):** all apt calls now go through `apt_noninteractive()`, which runs `sudo env DEBIAN_FRONTEND=noninteractive NEEDRESTART_MODE=a apt-get -y -q`. The top-level exports are gone; no direct `apt-get` calls remain. The upgrade keeps `--force-confdef/confold` on top.
- **Req 7 gap:** preflight now also checks `archive.ubuntu.com`, `registry.npmjs.org`, `nodejs.org` (all verified to answer 200 on HEAD).
- **Req 9 edge:** the GitHub-login announcement now triggers unless `~/.git-credentials` actually contains a `github.com` entry, not merely exists.
- **Req 15:** after the `[daemon]` fallback sed, the script re-checks the file and warns with manual instructions instead of claiming success.
- **Req 24:** the final summary only claims the screen won't blank when gsettings succeeded; on failure it adds a manual step instead.
- **`$USER` under `set -u`:** replaced with `REAL_USER="$(id -un)"`.
- **mktemp leak:** the TeamViewer temp dir is now removed, not just the deb.
- **chmod 666 → 644** on `/var/restart_lock` and the log (cron runs as root).

Left as-is, deliberately: the `chown -R /usr/local` (inherited behavior `n` needs) and the first-launch network need for the extension fetch (covered by the watchdog + the machine-#1 checklist).

PR #50, opened 2026-09-09
chore: pin dependencies and fold install into the startup script

zbyju:
Three changes that prepare the boxes for a real dependency update. **No package was upgraded** — versions are byte-for-byte what the lockfile already resolved.

## 1. Pin dependencies to exact versions

All dependencies across the 5 `package.json` files now use exact versions instead of caret ranges, with the lockfile `specifiers:` block updated to match. No resolution line changed.

This matters because `override()` copies backend's `package.json` into `dist` and runs `pnpm install` there **with no lockfile**. With carets, every box resolved whatever was newest-in-range at that moment, so boxes drifted apart from each other and over time. Now the install is deterministic.

Verified: `pnpm install --frozen-lockfile` → "Lockfile is up-to-date".

## 2. Do the full install from the startup script

`startup.sh` now does everything; the separate install path is gone.

New `source/apps/startup/versions.env` is the only place toolchain versions live:

```
NODE_VERSION=18.12.1
PNPM_VERSION=7.5.0
PM2_VERSION=6.0.14
N_VERSION=10.2.0
```

Values are bare and unquoted, because `startup.bat` parses this file by hand and understands nothing else. The header states that rule, and names the two other files that carry a copy of the pnpm version.

On every boot `startup.sh` compares and converges to these exact versions — installing or downgrading, never "latest". There is no `set -e`. Every step is best-effort and logs to `source/logs/startup.sh.log`, so a failed step never stops the panel from starting.

### node_modules are only wiped when they are already broken

A first draft wiped `node_modules` and retried whenever `pnpm install` failed. On a box with no network that means: install fails, modules deleted, retry fails, box cannot start at all. Now the clean reinstall only happens when the existing modules are already unusable — and the probe checks one package per app, since each app has its own `node_modules`.

If even the clean reinstall fails, `startup.sh` exits non-zero instead of starting a panel that has no `node_modules` to run on.

### Removed

`src/logic/install/{ubuntu,windows}.js`, `scripts/ubuntu/install.sh`, `scripts/windows/install.bat`, the `--install` flag, and the `dev:install` / `start:install` scripts. `sudo-prompt` is dropped from `startup`'s dependencies; only the deleted Windows installer used it.

`install-all.sh` hands off to `startup.sh` instead of duplicating the install, and its final summary now prints the versions the box actually has rather than its own constants.

`startup.sh` and `startup.bat` keep their exact paths and names, so the autostart entry and the Windows StartUp symlink still resolve.

## 3. Pull before reading versions.env

Review caught that boot-time convergence converged the *previous* boot. `versions.env` and the lockfile reach a box through `git pull`, which ran inside `node src/index.js` — after `startup.sh` had already sourced `versions.env` and run the frozen install. A Node or pnpm bump therefore landed one boot late, and `internet_check.sh` only reboots on Mondays, so "one boot late" can mean a week.

`startup.sh` now owns the whole sequence:

1. `git pull` in the repo root
2. re-exec itself once if the pull changed anything, so the newly pulled script runs — `BABYBOX_REEXEC` guards against a loop
3. source `versions.env`, converge node / pnpm / pm2 / n
4. `pnpm install --frozen-lockfile`, with the recovery above
5. `node src/index.js --ubuntu [--updated]`

`update()` is gone from both start modules. Because the shell now owns the pull, a second `git pull` inside node would always report "already up to date" and `build()` would never fire — so the shell passes `--updated` and `onStartup(updated)` takes it as a parameter.

Ordering matters beyond tidiness: a pnpm major bump means boot N would otherwise run `pnpm install --frozen-lockfile` with the old pnpm against a lockfile written by the new one. Converging first avoids that entirely.

## Testing

New `apps/startup/tests/startup.test.sh` — 53 assertions, run with `pnpm -F startup test`. It runs the real `startup.sh` against stub `node`, `npm`, `n`, `pnpm`, `pm2` and `git` in a throwaway `HOME`, so it installs and deletes nothing on the machine running it.

| Case | Asserts |
|---|---|
| Versions already match | nothing is installed |
| Versions all wrong | each one is installed, `pm2 update` runs |
| `n` at the wrong version | reinstalled to the pin |
| `n` install fails | `NODE_MISMATCH` logged, panel still starts |
| `versions.env` missing a key | no unpinned `npm install -g` |
| Install fails, deps work | `node_modules` kept |
| Install fails, only backend broken | clean reinstall |
| Install fails after cleaning | panel is not started, exit 1 |
| Pull finds nothing | no re-exec, no `--updated` |
| Pull finds a commit | exactly one re-exec, `--updated` passed |
| Pull reports a change every time | loop guard holds at one re-exec |
| Pull fails | panel still starts |

The tests are mutation-checked. Removing the re-exec, the loop guard, the `--updated` handoff, the per-app dependency probe, the `install_deps` exit, or the `n` version check each breaks 2–4 assertions. The harness has its own watchdog, so a runaway re-exec fails the test instead of hanging it.

Everything else: `pnpm install --frozen-lockfile` clean, panel builds, backend and configer `tsc` report no errors in repo files, eslint and prettier clean on the startup app. Backend jest 34/35 — the one failure (`url.test.ts`) is pre-existing and needs a live configer.

## Please check before merging

- **pm2 6.0.14 is my pick, not a considered fleet decision.** Boxes currently run whatever `pm2@latest` gave them on their provisioning date, so they differ today. 6.0.14 is the last release supporting Node >=16; pm2 7 needs >=18 and would also work on 18.12.1. Converge one box and confirm `pm2 update` keeps `babybox` and `configer` running before this reaches the fleet.
- **`startup.bat` is untested.** No Windows available. It mirrors the shell version, including the pull, the re-exec and the loop guard, and that is the trickiest control flow in the file. It needs a real run.

## Still open

Not touched here, and worth a follow-up:

- **Nothing runs `pnpm install --frozen-lockfile` in CI.** This is the gap that let #59 and #61 drift the lockfile, fixed after the fact by #64. It matters more once this lands, because `startup.sh` now exits non-zero on a failed frozen install rather than booting anyway.
- `build()` treats any stderr as failure ([ubuntu.js](https://github.com/zbyju/babybox/blob/main/source/apps/startup/src/logic/start/ubuntu.js)). An earlier draft of this description blamed pnpm deprecation warnings for this — that is wrong, pnpm 7.5.0 writes them to stdout, tested directly. The check is still fragile, because Node process warnings do go to stderr, but the specific mechanism claimed before does not apply.
- `git pull` on the boot path has no timeout. A box whose network hangs rather than fails can block panel startup. This is pre-existing — the pull was always in front of the panel start, just inside node.
- The rollback in `override()` uses `cwd: "../../dist"`; the correct path is `../../../dist`.
- `dist` installs devDependencies too — jest, newman, eslint, typescript.
- Node 18.12.1 is EOL (April 2025) and blocks every current major. Moving it is now a one-line change to `versions.env`, but should be its own PR.


source/apps/startup/scripts/ubuntu/startup.sh
```diff
@@ -1,2 +1,111 @@
...
+  run pnpm install --frozen-lockfile || { log "pnpm install selhal i po vycisteni"; return 1; }
+}
+
+# ----- Beh -------------------------------------------------------------------
+
+log "Start"
+
+if [ -f "$STARTUP_DIR/versions.env" ]; then
+  # shellcheck source=../../versions.env
+  . "$STARTUP_DIR/versions.env"
```
zbyju: `rv-doubt` ⚠️ **Design gate — the boot converges to the previous boot's `versions.env`.**

`versions.env`, the lockfile and every `package.json` reach a box through one `git pull`. That pull runs inside `node src/index.js --ubuntu`, which is started on line 111 — after this file is sourced and after `pnpm install --frozen-lockfile` has already run. So boot-time convergence never converges the boot it runs in. It converges the boot before.

Walk through the upgrade this PR prepares for, Node 18 to 20 and pnpm 7 to 9 in one commit:

1. Boot N sources the old `versions.env`, keeps Node 18 and pnpm 7.5.0, and runs a frozen install against the old lockfile. That result says nothing about the new code.
2. `git pull` brings the upgrade. `update()` returns Updated, so `build()` runs. The root `build` script starts with a plain `pnpm install`, not frozen, so pnpm 7.5.0 reads a lockfile written by pnpm 9.
3. That most likely errors. `build()` returns Error, `override()` is skipped, and the box keeps serving the old `dist`.
4. Boot N+1, up to a week later, installs Node 20 and pnpm 9. But `git pull` now says already up to date and `dist` exists, so `build()` is skipped.
5. `dist` was built by Node 18. Nothing rebuilds it until an unrelated commit arrives.

Checked in the code, not forecast:

- `src/logic/start/ubuntu.js:59` — `git pull` is the only place a box updates `versions.env` and the lockfile.
- `src/logic/start/ubuntu.js:231` — `build()` runs only when the pull changed something, or `dist` is missing.
- `source/package.json:16` — `"build": "pnpm install && ..."`, not frozen.
- `scripts/ubuntu/internet_check.sh:18` — weekly Monday reboot, so the next boot can be seven days away.
- `startup.bat:15-48` has the same order, so both platforms share this.

**Alternative.** Make the boot one pipeline with one owner. `startup.sh` runs `git pull` first, then sources `versions.env`, converges, re-execs itself once if the pull changed anything, then hands over to node with `update()` removed. Cost: `git pull` loses the winston logging in `update()`, and the re-exec needs a loop guard. About 20 lines per script, plus deleting `update()` and its two call sites.

**Smaller variant.** Keep provisioning out of the boot path. Leave it in `install-all.sh` where a person watches it, and let `onStartup()` — which already knows a commit arrived — converge and rebuild in the same boot. This also removes the network from every steady-state boot.

**Decision needed.** Confirm how a toolchain change is meant to reach a box, and whether one boot or two is acceptable. If two is acceptable, `build()` still needs a reason to run on the second boot, or `dist` stays stale.

The ordering is certain from the code. The pnpm 7 against pnpm 9 lockfile failure is reasoned, not observed.

source/apps/startup/scripts/ubuntu/startup.sh
```diff
@@ -1,2 +1,111 @@
...
+  ensure_node
+  ensure_pnpm
+  ensure_pm2
+else
+  log "versions.env chybi — kontrolu verzi preskakuji"
+fi
+
+cd "$STARTUP_DIR" || { log "Adresar $STARTUP_DIR neexistuje — koncim"; exit 1; }
+
+install_deps
```
zbyju: `rv-tests` · `rv-logic` · `rv-sweep` 🔴 `install_deps` returns 1 on total failure and nothing reads it, so the panel starts with no `node_modules`.

What happens on a box:

1. `pnpm install --frozen-lockfile` fails. A registry outage or a full disk is enough.
2. `deps_ok` also fails, so line 89 deletes `node_modules` for every app, including `apps/startup`.
3. The retry on line 90 fails for the same reason.
4. Line 111 runs `node src/index.js --ubuntu` anyway. Node throws on `require("winston")`, its own deleted dependency.
5. The panel never starts. `internet_check.sh` only pings the gateway, so no reboot recovers this state.

The same missing check hits provisioning. Before this PR, step 13 of `install-all.sh` ran `pnpm install` inline, so `set -e` and the ERR trap printed INSTALACE SELHALA. Now that work sits here, and `src/index.js` exits 0 even when it logs a start error. A technician sees INSTALACE DOKONCENA and leaves a dead box.

```suggestion
install_deps || { log "Zavislosti se nepodarilo nainstalovat - panel nespoustim"; exit 1; }
```

source/apps/startup/scripts/ubuntu/startup.sh
```diff
@@ -1,2 +1,111 @@
...
+
+  have="$(node -v 2>/dev/null)"
+  [ "$have" = "$want" ] || { log "Node je porad $have"; return 1; }
+  log "Node $want nainstalovan"
+}
+
+ensure_pnpm() {
+  [ "$(pnpm --version 2>/dev/null)" = "$PNPM_VERSION" ] && return 0
+  log "Instaluji pnpm $PNPM_VERSION"
+  run npm install -g "pnpm@$PNPM_VERSION" || { log "pnpm se nepodarilo nainstalovat"; return 1; }
```
zbyju: `rv-types` 🔴 An empty or missing key in `versions.env` installs the latest version, which this file forbids.

Nothing checks that the four keys exist. An absent key and an empty key are the same value here, and there is no `set -u`.

1. Someone renames or drops the `PNPM_VERSION` line.
2. `PNPM_VERSION` is empty. The box has pnpm 9, so the equality check on line 58 fails.
3. This line runs `npm install -g pnpm@`.
4. npm reads an empty spec as latest and installs it.

I checked step 4: `npm install --dry-run 'is-odd@'` resolves to 3.0.1, the latest.

The opposite hole is worse. If `PM2_VERSION` is empty and pm2 is missing, `pm2 --version` prints nothing, `[ "" = "" ]` passes on line 64, and `ensure_pm2` returns without installing pm2 at all. `startup.bat:34-45` has both holes.

All four keys are present today, so this needs a bad edit to fire. The guard is two lines, and the header of `versions.env` forbids exactly this outcome.

Fix, after sourcing on line 99:

```bash
for v in NODE_VERSION PNPM_VERSION PM2_VERSION N_VERSION; do
  [ -n "${!v}" ] || { log "$v chybi ve versions.env"; exit 1; }
done
```

source/apps/startup/scripts/ubuntu/startup.sh
```diff
@@ -1,2 +1,111 @@
...
+}
+
+ensure_pm2() {
+  [ "$(pm2 --version 2>/dev/null)" = "$PM2_VERSION" ] && return 0
+  log "Instaluji pm2 $PM2_VERSION"
+  run npm install -g "pm2@$PM2_VERSION" || { log "pm2 se nepodarilo nainstalovat"; return 1; }
```
zbyju: `rv-brief` 🟡 Every box installs pm2 6.0.14 and runs `pm2 update` on its next boot, with no canary.

pm2 runs the panel itself (`pm2 start ../dist/index.js -n babybox`). If pm2 6.x breaks that process, every box fails the same day.

The PR body already flags 6.0.14 as your pick rather than a fleet decision, and today's fleet version is unknown, so the size of the jump is unknown. The stub harness never ran a real daemon migration.

Fix: converge one box first and confirm `pm2 update` keeps `babybox` and `configer` running, before this reaches the fleet.

source/apps/startup/scripts/ubuntu/startup.sh
```diff
@@ -1,2 +1,111 @@
...
+  # Bezici demon zustane na stare verzi, dokud ho pm2 update nerestartuje
+  run pm2 update || log "pm2 update selhal"
+}
+
+# ----- Zavislosti ------------------------------------------------------------
+
+deps_ok() {
+  node -e "require('winston'); require('fs-extra')" >/dev/null 2>&1
```
zbyju: `rv-brief` 🟡 `deps_ok` tests only the startup app's own two packages, but its result decides for all four apps.

Each workspace app has its own `node_modules`, and this probe runs from `apps/startup`.

1. `backend/node_modules` is missing `express`.
2. `startup/node_modules` still has `winston` and `fs-extra`.
3. `deps_ok` returns true, so the clean reinstall is skipped.
4. The backend keeps failing until someone fixes it by hand.

Fix: check one package per app, or always retry `pnpm install --frozen-lockfile` once before giving up.

source/apps/startup/scripts/ubuntu/startup.sh
```diff
@@ -1,2 +1,111 @@
...
+
+  log "Node je $have, chceme $want"
+  if ! command -v n >/dev/null 2>&1; then
+    run npm install -g "n@$N_VERSION" || { log "n se nepodarilo nainstalovat"; return 1; }
+  fi
+  run n "$NODE_VERSION" || { log "n $NODE_VERSION selhalo"; return 1; }
+  hash -r
+
+  have="$(node -v 2>/dev/null)"
+  [ "$have" = "$want" ] || { log "Node je porad $have"; return 1; }
```
zbyju: `rv-tests` 🟡 When `n` fails, the box runs unpinned Node and no fleet check can find it.

Fail-soft is intended here, and that is the right call. The problem is the signal. This log line sits in the same stream as every other message, so nobody can list the boxes that did not converge.

Example: `NODE_VERSION` moves from 16 to 18 for a feature the panel needs. On one box the `n` mirror is briefly unreachable. That box runs `install_deps` and the panel under Node 16, and looks the same as a healthy box.

Fix: write one dedicated, greppable line on this path, for example `NODE_MISMATCH`, so a fleet check can find un-converged boxes.

source/apps/startup/scripts/ubuntu/startup.sh
```diff
@@ -1,2 +1,111 @@
...
+}
+
+ensure_node() {
+  local want="v$NODE_VERSION"
+  local have
+  have="$(node -v 2>/dev/null)"
+  [ "$have" = "$want" ] && return 0
+
+  log "Node je $have, chceme $want"
+  if ! command -v n >/dev/null 2>&1; then
```
zbyju: `rv-brief` 🟡 `n` is installed only when it is missing, so `N_VERSION` is never enforced.

`versions.env` pins `N_VERSION="10.2.0"` so this tool converges too, but an existing `n` at any version passes `command -v n` and the pinned install is skipped. A box with an old `n` then installs Node with a tool the repo does not control.

Fix: compare `n --version` against `N_VERSION` the way `ensure_pnpm` and `ensure_pm2` compare theirs, and reinstall on a mismatch.

source/apps/startup/versions.env
```diff
@@ -0,0 +1,11 @@
+# Pinned toolchain for the panel computers. startup.sh reads this on every boot
+# and installs these exact versions when the machine has something else.
+#
+# Never widen these to ranges or "latest". The boxes run unattended in
+# hospitals, so a new version must only ever arrive because someone changed
+# this file and shipped it.
+
+NODE_VERSION="18.12.1"
+PNPM_VERSION="7.5.0"
```
zbyju: `rv-types` · `rv-sweep` 🟡 The pnpm version lives in three files that do not reference each other.

This file, `install-all.sh:24`, and `source/package.json:7` (`"packageManager": "pnpm@7.5.0"`) all hold 7.5.0 on their own. The header above says a new version must only arrive by editing this file, but editing this file alone is not enough.

1. Someone sets `PNPM_VERSION` to 8.15.0 here and ships it.
2. `startup.sh` installs pnpm 8.15.0 on every box at boot.
3. `packageManager` still says `pnpm@7.5.0`.
4. pnpm 9 refuses to run on that mismatch. pnpm 8 rewrites the lockfile and the frozen check fails.

The lockfile is `lockfileVersion: 5.4`, which pnpm 8 rewrites to 6.0. `install-all.sh` at least documents that its copy is bootstrap only and heals afterwards. `packageManager` does not heal.

Fix: name `source/package.json` and `install-all.sh` in the header above as files that must move with this one, or compare `packageManager` against `$PNPM_VERSION` in `startup.sh` and log an error when they differ.

source/apps/startup/scripts/windows/startup.bat
```diff
@@ -1,2 +1,67 @@
...
+
 cd ../../
-pnpm run start
+
+if not exist versions.env (
+  echo versions.env chybi - kontrolu verzi preskakuji
+  goto deps
+)
+
+for /f "usebackq eol=# tokens=1,2 delims==" %%a in ("versions.env") do (
```
zbyju: `rv-types` 🟡 This parser accepts less than bash does, so a valid edit to `versions.env` drops keys on Windows only.

`startup.sh` sources the file as bash. This reads a narrow `KEY=VALUE` grammar by hand, so the two platforms disagree about what the file means.

`export PNPM_VERSION="8.15.0"` is valid bash. Here the split on `=` makes token 1 `export PNPM_VERSION`, no `if` matches, and the value stays empty. That feeds straight into the empty-spec problem: `npm install -g pnpm@` installs the latest.

`%%~b` also strips only double quotes. `PNPM_VERSION='8.15.0'` becomes the literal `pnpm@'8.15.0'` and the install fails.

Fix: restrict `versions.env` to bare `KEY=value` with no quotes and no `export`, and state that rule in its header.

source/apps/startup/scripts/windows/startup.bat
```diff
@@ -1,2 +1,67 @@
...
+REM
+REM Zastupce ve slozce StartUp odkazuje na tento soubor. Nesmi se presunout
+REM ani prejmenovat, jinak se panel nespusti.
+REM
+REM Skript nejdriv srovna verze nastroju podle versions.env a pak spusti panel.
+REM Zadny krok nesmi spusteni zastavit - chyby se jen vypisou a pokracuje se
+REM s tim, co uz na pocitaci je.
+REM ============================================================================
+
 cd ../../
```
zbyju: `rv-logic` 🟡 This `cd` is unchecked, and the recursive deletes on lines 61-62 are relative to it.

`cd` in cmd does not change drive and does not stop the script. The wrong-directory case is exactly the case that reaches the delete: when the cwd is not the startup app, both `pnpm install --frozen-lockfile` and the `winston` probe fail, so the script falls through to `rmdir /s /q ..\..\node_modules` and then deletes `node_modules` inside every sibling folder.

`startup.sh` does not have this problem. It deletes absolute paths under `$BABYBOX_DIR` and guards its `cd` with `|| exit 1`.

```suggestion
cd /d "%~dp0..\.."
if errorlevel 1 exit /b 1
```

source/apps/startup/scripts/windows/startup.bat
```diff
@@ -1,2 +1,67 @@
...
+
+for /f "usebackq eol=# tokens=1,2 delims==" %%a in ("versions.env") do (
+  if "%%a"=="NODE_VERSION" set NODE_VERSION=%%~b
+  if "%%a"=="PNPM_VERSION" set PNPM_VERSION=%%~b
+  if "%%a"=="PM2_VERSION" set PM2_VERSION=%%~b
+)
+
+REM Windows nema spravce verzi Node (na Ubuntu to resi 'n'), takze verzi jen
+REM hlasime a instalaci musi udelat clovek.
+for /f "delims=" %%v in ('node -v 2^>nul') do set HAVE_NODE=%%v
```
zbyju: `rv-logic` 🟡 `for /f` sets nothing when the command produces no output, so the warning prints the variable name.

With no Node installed, `HAVE_NODE` stays undefined and cmd expands `%HAVE_NODE%` to that literal text. The console shows `POZOR: Node je %HAVE_NODE%, ma byt v18.12.1`. The branch is still taken, so only the message is wrong.

`HAVE_PNPM` and `HAVE_PM2` have the same hole. Their values are not printed, so they cost nothing today.

Also add `setlocal` after `@echo off`. Without it these three variables survive from the parent environment when the script runs twice in one cmd session.

```suggestion
set "HAVE_NODE=none"
for /f "delims=" %%v in ('node -v 2^>nul') do set HAVE_NODE=%%v
```

source/apps/startup/scripts/ubuntu/install-all.sh
```diff
@@ -458,15 +459,8 @@ fi
...
-  npm install -g "pnpm@$PNPM_VERSION"
-fi
-cd "$BABYBOX_DIR/source/apps/startup"
-info "Instaluji zavislosti (pnpm install)..."
-pnpm install
-info "Spoustim instalaci a prvni start panelu..."
-node src/index.js --install --ubuntu
+info "Zavislosti a verze nastroju resi startup.sh podle versions.env."
+"$UBUNTU_SCRIPTS_DIR/startup.sh"
 ok "Panel nainstalovan a spusten"
```
zbyju: `rv-sweep` 🟡 Really about `install-all.sh:476`: the final summary reports a pnpm version this script no longer installs.

This diff removed the only real use of `PNPM_VERSION` from line 24. It now appears once, in `echo "  - Node.js $NODE_VERSION, pnpm $PNPM_VERSION"`. `startup.sh` installs pnpm from `versions.env` instead, so once those two disagree the technician is told a version the box does not have.

The summary also never mentions pm2, which every box now installs.

Fix: print `node -v`, `pnpm --version` and `pm2 --version` in the summary instead of the constants.

source/apps/startup/package.json
```diff
@@ -5,24 +5,21 @@
   "main": "src/index.js",
   "scripts": {
     "dev": "nodemon src/index.js",
```
zbyju: `rv-sweep` 🟡 `nodemon` is not declared anywhere, and the deleted `installDeps()` was the only thing installing it.

`pnpm dev` here runs `nodemon src/index.js`. pnpm looks in `apps/startup/node_modules/.bin` and `source/node_modules/.bin`, and `nodemon` is in neither. `apps/configer` breaks the same way with `nodemon --ignore '*.json' --esm ./src/index.ts`.

I checked the other globals the deleted `installDeps()` covered. `typescript` and `ts-node` are pinned in the root `package.json`, so workspace scripts still find them. `nodemon` is declared only in `apps/backend`.

This is dev only. No panel computer is affected.

Fix: add `"nodemon": "2.0.19"` to the `devDependencies` of `apps/startup` and `apps/configer`, matching the pin in `apps/backend`.

source/apps/startup/scripts/ubuntu/startup.sh
```diff
@@ -1,2 +1,111 @@
-cd ~/babybox/source/apps/startup/
-pnpm run start --ubuntu
+#!/bin/bash
```
zbyju: `rv-tests` 🟡 The stub harness that verified this script is not in the repo, so nobody can rerun it.

The PR body describes four cases run against stubbed `node`, `npm`, `n`, `pnpm` and `pm2` binaries in an isolated tree. That work is the only check on 111 new lines of boot logic, and the next person to touch this file starts from nothing. The startup app has jest as a devDependency and zero test files, before and after this PR, so there is no pattern to follow either.

Worth checking in only if it covers the transitions rather than the happy path: version already matches so no install runs, version differs so an install runs, the install command itself fails, and the three `install_deps` outcomes.

Fix: add one bats or plain bash file under `apps/startup` that runs `startup.sh` against a fake PATH of stub binaries and asserts on the log lines for each branch above.

source/apps/startup/scripts/ubuntu/startup.sh
```diff
@@ -1,2 +1,111 @@
...
+  "$@" >>"$LOG_FILE" 2>&1
+}
+
+# ----- Kontrola verzi --------------------------------------------------------
+
+ensure_npm_prefix() {
+  local want="$HOME/.npm-global"
+  [ "$(npm config get prefix 2>/dev/null)" = "$want" ] && return 0
+  mkdir -p "$want"
+  run npm config set prefix "$want" || log "npm prefix se nepodarilo nastavit"
```
zbyju: `rv-hygiene` 🟢 This returns 0 on failure while `ensure_node`, `ensure_pnpm` and `ensure_pm2` all return 1.

No caller reads any of these return values today, so nothing breaks. It is a consistency nit, not a bug.

```suggestion
  run npm config set prefix "$want" || { log "npm prefix se nepodarilo nastavit"; return 1; }
```

source/apps/startup/src/index.js
```diff
@@ -2,59 +2,31 @@ const winston = require("winston");
...
 const ubuntuStart = require("./logic/start/ubuntu");
-const winInstall = require("./logic/install/windows");
-const ubuntuInstall = require("./logic/install/ubuntu");
 const { getFulltimeFormatted } = require("./utils/time");
 
 async function main() {
-  let canStartup = true;
   const args = process.argv.slice(2);
-  const shouldInstall = args.find((a) => a.toLowerCase() === "--install");
   const isUbuntu = args.find((a) => a.toLowerCase() === "--ubuntu");
```
zbyju: `rv-types` 🟢 `isUbuntu` holds `"--ubuntu"` or `undefined`, so the name does not match the value.

No suggestion block, because line 21 has to change with it:

```js
const isUbuntu = args.some((a) => a.toLowerCase() === "--ubuntu");
// line 21
const res = isUbuntu ? await ubuntuStart() : await winStart();
```

zbyju:
## Review — 16 findings, all inline on the diff

🔴 2 look broken · 🟡 11 worth a look · 🟢 3 nits · ⚠️ 1 design gate on `startup.sh:99`

Seven review agents ran: `rv-brief`, `rv-hygiene`, `rv-doubt`, `rv-logic`, `rv-types`, `rv-tests`, `rv-sweep`. Each inline comment carries the tag of the agent that raised it.

### WHAT
Every dependency in the five `package.json` files moves from a caret range to an exact version, and the lockfile `specifiers:` block matches. Separately, the one-time install scripts are gone. `startup.sh` and `startup.bat` now converge Node, pnpm and pm2 to the versions in a new `versions.env`, then run the dependency install.

### RISK — medium
- **Blast radius** — `versions.env` is one file for the whole fleet. Every box installs pm2 6.0.14 and runs `pm2 update` on its next boot, with no canary.
- **External coupling** — convergence needs the npm registry and the `n` mirror at boot. Failures are logged and swallowed, so a box can stay unconverged and nobody sees it.
- **Reversibility** — no evidence that `pm2 update` is undone by putting an older pm2 version back in `versions.env`.

### WHY
`override()` copies backend's `package.json` into `dist` and runs `pnpm install` there with no lockfile. Under caret ranges each box resolved whatever was newest at that moment, so boxes drifted apart. Exact pins make that install deterministic. Folding the install into `startup.sh` means a box with a stale toolchain fixes itself instead of needing a person on site.

### HOW
```
startup.sh (boot)
  -> source versions.env                                     new file
  -> ensure_npm_prefix()  ~/.npm-global, so no sudo           new
  -> ensure_node()        n $NODE_VERSION on mismatch         new
  -> ensure_pnpm()        npm i -g pnpm@$PNPM_VERSION         new
  -> ensure_pm2()         npm i -g pm2@$PM2_VERSION, pm2 update  new
  -> install_deps()       pnpm install --frozen-lockfile
       on failure -> deps_ok(): require('winston'), require('fs-extra')
       ok -> keep node_modules;  broken -> rm -rf, reinstall
  -> node src/index.js --ubuntu        unchanged entry point
No `set -e`. Every failure is logged and swallowed.
```

### TOUCHES

**Dependencies** — 81 pins across 5 manifests, `sudo-prompt` removed. I checked the lockfile diff line by line: only `specifiers:` entries and the `sudo-prompt` block changed. No `version:` line moved. The pin-only claim in the description holds.

**Tests** — 0 added, 0 changed.
- ✓ `startup.sh` run by hand against stubbed `node`/`npm`/`n`/`pnpm`/`pm2`: all versions wrong, all correct, install fails with working deps, install fails with broken deps
- ✓ backend, configer and panel build. eslint clean on the startup app. Backend 30/31, and `url.test.ts` fails on `main` too
- ✗ `startup.bat` — no Windows machine, as the description says
- ✗ a real `pm2 update` across a major version against a live daemon
- ? the stub harness is not in the repo, so none of the results above can be reproduced

### One finding that has no line on the diff

🟢 **The determinism claim covers direct dependencies only** — `rv-brief`

`override()` in `src/logic/start/ubuntu.js:139-143` copies backend's `package.json` into `dist` and runs `pnpm install` there, but it still copies no lockfile. Direct dependencies are pinned now, so those are fixed. Transitive versions in `dist` can still differ between boxes and over time.

Fix: copy the lockfile into `dist` as well, or soften the claim in the description to direct dependencies.

That file is not part of this diff, so this is a note, not a change request.

source/apps/startup/scripts/ubuntu/startup.sh
zbyju: **Addressed in** [`ae08d67`](https://github.com/zbyju/babybox/commit/ae08d67)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- `install_deps` failure now exits 1
- Panel is not started with no node_modules
- `install-all.sh` ERR trap fires again
- Same guard added to `startup.bat`
- Covered by a new test case

source/apps/startup/scripts/ubuntu/startup.sh
zbyju: **Addressed in** [`dbb739d`](https://github.com/zbyju/babybox/commit/dbb739d)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- New `require_versions` checks all four keys
- Same guard in `startup.bat`
- Covered by a new test case

One change from the suggestion: on a bad `versions.env` the script skips version convergence and logs, instead of `exit 1`. An exit would brick every box in the fleet on one bad edit, which is worse than the hole it closes, and it contradicts the file header. The actual hole — `npm install -g pnpm@` resolving to latest — is closed either way.

source/apps/startup/scripts/ubuntu/startup.sh
zbyju: **Addressed in** [`360a8c3`](https://github.com/zbyju/babybox/commit/360a8c3)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- `deps_ok` probes one package per app
- startup/winston, backend/express, configer/express, panel/vue
- Same probe list in `startup.bat`
- Covered by a new test case

source/apps/startup/scripts/ubuntu/startup.sh
zbyju: **Addressed in** [`ad695d9`](https://github.com/zbyju/babybox/commit/ad695d9)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- Every un-converged Node path logs `NODE_MISMATCH`
- One greppable token, three failure paths
- Covered by a new test case

source/apps/startup/scripts/ubuntu/startup.sh
zbyju: **Addressed in** [`ad695d9`](https://github.com/zbyju/babybox/commit/ad695d9)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- New `ensure_n` compares `n --version` to `N_VERSION`
- Reinstalls on a mismatch
- Falls back to the old `n` if that install fails
- Covered by a new test case

source/apps/startup/versions.env
zbyju: **Addressed in** [`7dc99de`](https://github.com/zbyju/babybox/commit/7dc99de)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- Header now names both other copies
- `source/package.json` `packageManager` and `install-all.sh`
- States they must move together

Took the documentation option, not the runtime check. A boot-time compare would parse `package.json` in bash and only log, so it would not stop the mismatch either.

source/apps/startup/scripts/windows/startup.bat
zbyju: **Addressed in** [`7dc99de`](https://github.com/zbyju/babybox/commit/7dc99de)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- Values are now bare, unquoted
- Header states the format rule
- No quotes, no `export`, no spaces
- `%%~b` changed to `%%b`

source/apps/startup/scripts/windows/startup.bat
zbyju: **Addressed in** [`0cee1ed`](https://github.com/zbyju/babybox/commit/0cee1ed)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- Took the suggestion as written
- Path now comes from the script, not cwd

source/apps/startup/scripts/windows/startup.bat
zbyju: **Addressed in** [`ce3ea88`](https://github.com/zbyju/babybox/commit/ce3ea88)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- All three variables default to `zadny`
- Added `setlocal` after `@echo off`

source/apps/startup/scripts/ubuntu/install-all.sh
zbyju: **Addressed in** [`afdfe11`](https://github.com/zbyju/babybox/commit/afdfe11)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- Summary prints `node -v`, `pnpm --version`, `pm2 --version`
- pm2 is now in the summary
- Falls back to `?` when a tool is missing

source/apps/startup/package.json
zbyju: **Addressed in** [`4ef8a9b`](https://github.com/zbyju/babybox/commit/4ef8a9b)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- Added `nodemon@2.0.19` to startup and configer
- Lockfile regenerated with pnpm 7.5.0
- No resolution line changed, 7 added
- `--frozen-lockfile` still passes

source/apps/startup/scripts/ubuntu/startup.sh
zbyju: **Addressed in** [`5bef7db`](https://github.com/zbyju/babybox/commit/5bef7db)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- New `apps/startup/tests/startup.test.sh`, 33 assertions
- Stub node, npm, n, pnpm, pm2 in a throwaway HOME
- Covers version match, mismatch, failed install
- Covers all three `install_deps` outcomes
- Run with `pnpm -F startup test`

Checked the tests fail on the real bugs: reverting the `deps_ok` fix, the `install_deps` exit and the `n` version check each breaks two or three assertions.

source/apps/startup/scripts/ubuntu/startup.sh
zbyju: **Addressed in** [`b18cad4`](https://github.com/zbyju/babybox/commit/b18cad4)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- Took the suggestion as written

source/apps/startup/src/index.js
zbyju: **Addressed in** [`7509059`](https://github.com/zbyju/babybox/commit/7509059)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- `find` changed to `some`
- Line 21 now tests the boolean directly

source/apps/startup/scripts/ubuntu/startup.sh
zbyju: **Addressed in** — no commit
**Status:** not done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- This is a rollout step, not a code change
- One box must converge and be watched first
- The PR body already flags 6.0.14 as unverified

source/apps/startup/scripts/ubuntu/startup.sh
zbyju: **Addressed in** — no commit
**Status:** not done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- The boot order is confirmed, the finding stands
- Both alternatives need a decision first
- One boot or two is not recorded anywhere
- Left for a human to pick

zbyju: **Addressed in** [`7e174f1`](https://github.com/zbyju/babybox/commit/7e174f1)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- Took the first alternative, decided by a human
- `startup.sh` runs `git pull` before sourcing `versions.env`
- Re-execs once when the pull changed anything
- `BABYBOX_REEXEC` guards against a loop
- `update()` removed from both start modules
- Shell passes `--updated` so `build()` still fires
- Mirrored in `startup.bat`
- 20 new assertions, 53 total

The pull now happens before anything reads `versions.env`, so a Node or pnpm bump converges in the boot that pulls it, not the next one. Tests cover the ordering, the single re-exec, the loop guard, and a failed pull.

PR #55, opened 2026-09-09
chore: remove stale yarn.lock

zbyju:
Deletes the leftover yarn lockfile. The repo is pnpm-only: package.json sets `packageManager` to pnpm@7.5.0, and there is a pnpm-workspace.yaml plus a current pnpm-lock.yaml.

A second lockfile from another package manager is misleading and can be picked up by tooling or CI.

Verification: grepped all tracked files for `yarn` (case-insensitive). The only hits are two ignore patterns in the panel .gitignore (`yarn-debug.log*`, `yarn-error.log*`) — dead boilerplate, left alone as out of scope. No script, CI config, or docs invoke yarn; the Travis config installs and runs pnpm. The lockfile itself was last touched by the commit that moved sources into a separate dir.


zbyju (commented):
Verdict: correct, in scope, and safe to merge. No blocking findings.

Checks I ran on the PR branch:

- `git grep -i yarn` over all tracked files: only `source/apps/panel/.gitignore` lines 5-6 (`yarn-debug.log*`, `yarn-error.log*`). Matches the body. Nothing else reads the deleted path.
- Nothing in `source/apps/startup/scripts/*` (Ubuntu and Windows), `source/apps/backend/.travis.yml`, `source/apps/configer/scripts/run.bat`, or any package.json script invokes yarn or a lockfile path. Travis installs `pnpm` and runs `pnpm install`; `install-all.sh` pins `pnpm@7.5.0`. No Dockerfiles or GitHub workflows exist in the repo.
- Diff is exactly one deleted file, 513 lines. No unrelated edits.
- Body claim about history holds: `source/yarn.lock` was last touched by `590a51a` ("Move source files into separate dir", 2022-08-03).
- Commit hygiene: single commit `chore: remove stale yarn.lock`, no body, no trailers. Conforms.

- **note** — `source/apps/panel/pnpm-lock.yaml` is the same kind of leftover from that same 2022 commit. `apps/panel` is a workspace member (`pnpm-workspace.yaml` lists `apps/*`) and has an `apps/panel:` importer in the root `source/pnpm-lock.yaml`, so the nested one is dead and pnpm ignores it. Same argument as this PR. Out of scope here; worth a follow-up.

PR #56, opened 2026-09-09
fix: pass watchdog timeout as axios config not request body

zbyju:
`updateWatchdog` called `axios.put(url, { timeout })`. The second argument of `axios.put` is the request body, not the config.

So the panel sent `{"timeout":5000}` as the JSON body and the request ran with axios' default timeout, which is no timeout at all. A hung PUT would never return, and the panel loop's watchdog call could stall forever.

The watchdog is what stops the engine unit rebooting an unattended machine, so a stalled call has real consequences on site.

Fix: pass `null` as the body and `{ timeout }` as the config. The backend handler for `PUT /engine/watchdog` reads nothing from the request body, so dropping the body changes no behaviour.

Verification: read the backend route handler to confirm it ignores the body, and re-read the diff by hand. No local typecheck ran - `source/apps/panel/node_modules` is not installed and `pnpm install` was skipped as too slow.


zbyju (commented):
**Verdict: correct. Nothing blocking — one small note on `null` vs `undefined`.**

I read axios 0.27.2 itself (extracted from the local pnpm store, since `node_modules` is not installed) rather than going from memory.

**What I verified**

- `lib/core/Axios.js:143` — `function httpMethod(url, data, config)`. The old call put `{ timeout }` in the `data` slot and left `config` undefined, so the request ran with axios' default of no timeout. Confirmed.
- `lib/core/mergeConfig.js` — `timeout` merges via `defaultToConfig2`, so the value from the config argument survives. The fix does set the timeout.
- Backend ignores the body. `routes/engineRoute.ts:33` calls `updateWatchdog()` with no arguments; `fetch/fetchFromUnits.ts:119` builds the unit URL from config only. Neither touches `req.body`. `express.json()` in `index.ts` is the only body parser and it only fires on `application/json`, so dropping the JSON body just means it no longer parses. No handler notices.
- The whole path is now bounded: the backend's own hop uses `fetchFromUrl(url)` with its 5000 ms default (`fetch/fetchFromUrl` → `axios.get(url, { timeout })`), so panel-side timeout plus backend-side timeout close the hazard rather than half of it.
- Diff is one file, +1/-1. Commit is `fix: pass watchdog timeout as axios config not request body` — subject only, no body, no trailers. Matches convention.

**Findings**

- **note** — `null` vs `undefined`. In 0.27.2 `defaults.headers.put` is `{'Content-Type': 'application/x-www-form-urlencoded'}` (`lib/defaults/index.js:143`), and the xhr adapter deletes Content-Type only when `typeof requestData === 'undefined'` (`lib/adapters/xhr.js:159`). `typeof null === 'object'`, so `null` sends an empty body but still sends that Content-Type header; `undefined` would send neither. `null` does correctly avoid the JSON path — `isObject(null)` is false (`lib/utils.js:108`), so no `application/json` header and no stringified body. The stray header is harmless here because nothing on the backend reads a urlencoded body, so this is a preference, not a defect. If you want the request to carry no content headers at all, use `undefined`; `axios.put`'s `data` parameter is optional so it typechecks. Either is fine to merge.

- **note** — no other instance of this mistake in the panel. The only other write is `sendSettings` (`api/units.ts:104`), which already uses the three-argument form. Every remaining axios call in `apps/panel/src` is `get(url, config)`.

No test added, but the panel has no test script or test files, so that is not a gap this PR creates.

PR #57, opened 2026-09-09
fix: make RESTART_ENABLED able to disable auto-restart

zbyju:
Three fixes so the backend auto-restart kill switch actually works.

**1. `RESTART_ENABLED` could never turn auto-restart off**

`process.env.RESTART_ENABLED || true` is always truthy, and env values are strings anyway, so even `"false"` passed. The flag is now resolved by a predicate that treats `false`, `0`, `no` and `off` as disabled.

**2. `.env` was read too late for the flag to matter**

`export const modules = modulesObject()` runs at module load in `index.ts`, but `dotenv.config()` was the first line of `main()`, which runs after. So `RESTART_ENABLED`, `RESTART_INTERVAL` and `RESTART_ERROR_THRESHOLD` were all read before the `.env` file was parsed. Fixing point 1 alone would not have helped: an operator adding `RESTART_ENABLED=false` to `source/apps/backend/.env` would get no change and no warning, because `.env` is the only place this repo configures the backend.

`dotenv.config()` now runs at module scope above `modulesObject()` and is no longer called inside `main()`.

**3. Only the exact lowercase `"false"` disabled restarts**

`0`, `FALSE`, `No`, `off` all left reboots on. The value is now trimmed and lowercased before the check. Unset still means enabled, so nothing that is running today changes.

This matters because the panel runs unattended in a hospital and reboots its own host machine. Without a working kill switch there is no way to stop a reboot loop remotely during maintenance.

`RESTART_INTEVAL` was also missing an R and is renamed to `RESTART_INTERVAL`. Nothing in the repo sets either spelling, so the 20000 ms default is what runs today and the rename changes no live behaviour.

**Verification**

- Installed backend deps and ran `tsc -p tsconfig.json`. No errors in `src`. The only compiler output is unrelated parse noise from a `bun-types` package in a parent `node_modules`, present before this change.
- Ran the compiled `dist/index.js` with a temporary `.env` holding `RESTART_ENABLED=OFF` and `RESTART_INTERVAL=99000`, with a probe inside `modulesObject()`. It resolved `restartEnabled=false` and saw both values, so the `.env` route now works. Probe removed afterwards.
- Circular import checked. `restart.ts` imports `config` from `".."`. The emitted `restart.js` accesses it as `__1.config` inside `stopRestart`/`startRestart` only, never at module load, so the partial exports object during the cycle is fine. `require` calls are hoisted above every statement, so moving `dotenv.config()` does not change module init order at all. Running `node dist/index.js` before and after the change behaves identically.
- Grepped `process.env` across `source/apps/backend/src`. Every read other than the three restart vars sits inside a function called from `main()` or a route handler, all after `dotenv.config()`. `checkInit.ts` reads env but has no callers.
- `NODE_ENV`, `PORT` and `API_PREFIX` are unaffected: they are read inside `main()`, which still runs after `dotenv.config()`, just later than it used to.

**Left as follow-ups**

- `restartRepository()` still registers its `setInterval` regardless of the flag; disabling works because `lastRequest` stays `null`. Passing the flag in and skipping the timer is a real design fix but a bigger change than this PR.
- No test covers `restart.ts` or `init.ts`. There is no existing harness for these modules, so adding one is its own change.


zbyju (commented):
Both one-liners are correct and the default is preserved, but the kill switch still cannot be turned off through `.env` — env is read before `dotenv.config()` runs. One blocking fix needed.

**blocking — `RESTART_ENABLED=false` in `.env` has no effect**

`source/apps/backend/src/index.ts:17` runs `modulesObject()` at module load. `dotenv.config()` is the first line of `main()` (line 23), which runs after. So `process.env.RESTART_ENABLED` is read before the `.env` file is parsed, and the same goes for `RESTART_INTERVAL` and `RESTART_ERROR_THRESHOLD` in `restart.ts`.

`.env` is the only place this repo configures the backend (`PORT`, `NODE_ENV`, `API_PREFIX` all live there). Whoever needs to stop a reboot loop remotely will add `RESTART_ENABLED=false` to `source/apps/backend/.env`, restart, and get no change and no warning. Only a real OS-level env var works today.

Fix: move `dotenv.config()` above line 17 in `index.ts`, i.e. to module top level before `export const modules = modulesObject();`. One line, keeps the PR small. If you would rather not touch `index.ts` here, say in the PR and the README that the var must be exported by the shell/service, not put in `.env` — but the code fix is cheaper and less likely to be forgotten.

**non-blocking — only the exact string `"false"` disables**

`"0"`, `"FALSE"`, `"No"`, `"off"` and `""` all leave auto-restart on. This is a fail-open safety switch: a wrong spelling looks like it worked and the machine still reboots itself. There is no existing boolean-env convention in this repo to be consistent with, so pick the forgiving one:

```ts
const raw = (process.env.RESTART_ENABLED ?? "").trim().toLowerCase();
const restartEnabled = !["false", "0", "no", "off"].includes(raw);
```

Also log the resolved value once at startup. Right now nothing tells an operator whether the switch took effect, which is the thing they most need to confirm during maintenance.

**non-blocking — the switch works by accident, not by design**

`restartRepository()` is called unconditionally and registers its `setInterval` regardless of the flag. `restartEnabled` only gates `onIncomingRequest()`. Disabling works purely because `lastRequest` stays `null` and the interval hits `if (lastRequest === null) return;` on every tick. The timer runs forever doing nothing.

That holds today — `onIncomingData` is the only caller and `restartRoute` is its only consumer — so no reboot can happen with the flag off. But anyone who later calls `restartRepo.onIncomingRequest()` from somewhere else, or removes the null guard, silently re-arms the reboot. Follow-up, not this PR: pass the flag into `restartRepository()` and skip registering the interval when it is off.

Related, and pre-existing: the null guard means the watchdog never arms if the panel has never sent a single request since boot. Probably intended, worth a comment.

**note — rename verified safe**

Grepped the whole tree (excluding `node_modules`/`.git`) for `RESTART_INTEVAL`, `RESTART_INTERVAL`, `RESTART_ENABLED`, `RESTART_ERROR_THRESHOLD`, plus a case-insensitive sweep of every `.sh`, `.bat`, `.json`, `.md`, `.yml`, `.ts`, `.js`. Only hits are the three reads in the backend source. Both `.env` files hold just `NODE_ENV`/`PORT`/`API_PREFIX`; `configer/configs/base.json` and `versions.json` set nothing; the `startup` app and its ubuntu/windows scripts set nothing. So the interval is 20000 ms before and after. No live behaviour change.

**note — default preserved**

Unset: `undefined !== "false"` is `true`, same as `undefined || true`. Everything that was enabled stays enabled; only the literal `"false"` flips, which is the bug. Return type also tightens from `string | boolean` to `boolean`.

**note — scope and hygiene**

Diff is exactly the two intended lines. Commit `fix: make RESTART_ENABLED able to disable auto-restart` is subject-only, no body, no trailers — matches the convention. Two small things: the PR body carries a "Generated with Claude Code" line, and the title mentions only `RESTART_ENABLED` while the commit also carries the `RESTART_INTERVAL` rename.

**note — no test covers this**

`jest.config.json` exists but tests only live under `src/types/__tests__` and `src/utils/__tests__`. Nothing exercises `restart.ts` or `init.ts`. Fine for two lines, but on a path that reboots a hospital machine a small test that `RESTART_ENABLED=false` leaves `onIncomingRequest` uncalled would be worth adding once the dotenv ordering is fixed.

zbyju:
Both points addressed, two commits added.

**Blocking one — `.env` read too late.** `dotenv.config()` moved to module scope in `index.ts`, above `export const modules = modulesObject()`, and removed from `main()` so it runs once. Checks I ran:

- The cycle is safe. `restart.ts` reads `config` from `".."` only inside `stopRestart`/`startRestart`, which the emitted JS compiles to `__1.config` at call time, not at module load. `require` calls are hoisted above every statement anyway, so moving a statement cannot change module init order. `node dist/index.js` behaves the same before and after.
- Grepped `process.env` across `source/apps/backend/src`. Apart from the three restart vars, every read is inside a function reached from `main()` or a route handler, so all now after dotenv. `checkInit.ts` reads env but nothing calls it.
- `NODE_ENV`, `PORT`, `API_PREFIX` unchanged — still read inside `main()`, which still runs after the dotenv call.
- Proved it end to end: built `dist`, put `RESTART_ENABLED=OFF` and `RESTART_INTERVAL=99000` in a temporary `.env`, added a probe in `modulesObject()`. It resolved `restartEnabled=false` and saw both values. Probe and temp `.env` removed.

**Non-blocking one — brittle predicate.** Value is now trimmed, lowercased and matched against `false`, `0`, `no`, `off`. Unset still means enabled.

Build: installed backend deps and ran `tsc -p tsconfig.json`. Clean for `src`. The only output is unrelated parse noise from a `bun-types` package in a parent `node_modules`, which predates this branch.

Deliberately left out:

- Restructuring `restartRepository()` to take the flag and skip the `setInterval`. Agreed it is the right shape, but it is a bigger change than this PR.
- The startup log line for the resolved flag.
- Tests. There is no harness for these modules, so that is its own change.
- The stale `lastRequest`/`errorStreak` values in the returned object.

zbyju:
Follow-up in #65: the predicate fixed here was correct, but the env vars are read at module scope before `dotenv.config()` ran, so `RESTART_ENABLED` could not actually be set through `.env`.

PR #58, opened 2026-09-09
refactor: remove dead code from backend and panel loop

zbyju:
Removes code that nothing reads or calls.

- `source/apps/backend/src/utils/checkInit.ts` — whole file; `checkInit()` had no callers.
- `transformEngineData()` in `source/apps/backend/src/utils/transformData.ts` — no callers. Also dropped the `EngineData` import, which became unused. `transformThermalData()` stays.
- Dead `query` object in `source/apps/backend/src/routes/engineRoute.ts` — `query.timeout` was never read and `query.raw` just copied the existing `raw` const. The code now uses `raw` directly.
- Same dead `query` object in `source/apps/backend/src/routes/thermalRoute.ts`.
- `i = tryNumber;` in `updateSettings` in `source/apps/backend/src/fetch/fetchFromUnits.ts` — assigned to a local right before returning; `i` is never read after that.
- Unused `timeout = 5000` parameter on `updateWatchdogEngine()` in `source/apps/panel/src/logic/panel/panelLoop.ts` — never used in the body, and the only caller passes nothing.

No behaviour changes. Every removed item was confirmed unused by grepping the whole repo for each symbol before deleting it, and each remaining reference was checked by reading the surrounding code.

Verification: grepped the repo (excluding `node_modules`) for `checkInit`, `transformEngineData`, `EngineData`, `query.`, and `updateWatchdogEngine`, and read each call site. No `node_modules` is installed in this worktree for either app, so no build or typecheck was run — as instructed, `pnpm install` was skipped and the diff was re-read line by line instead.

Two things noticed but deliberately left alone:

- `engineRoute.ts` calls `transformThermalData` on engine data. Out of scope here.
- `thermalRoute.ts`'s `else` branch sends a `data` key while `engineRoute.ts`'s does not. Out of scope here.
- The `EngineData` type in `source/apps/backend/src/types/data.types.ts` now has no users. Left in place.


zbyju (commented):
Verdict: the "no behaviour change" claim holds. All six removals independently verified as dead. No blocking findings.

I re-grepped the whole tree (excluding `node_modules`/`.git`) on the PR head rather than trusting the description.

**Verified dead — all six**

1. `checkInit` — zero hits for the string anywhere in the repo after deletion: no import, no `require`, no dynamic path, no barrel (`src/utils/` has no `index.ts`), no test.
2. `transformEngineData` — zero hits. It also does not change which function runs on engine data: `engineRoute.ts` already imported and called `transformThermalData` on `origin/main`, before this PR. The import line is untouched. Only the `EngineData` import was dropped, and `ThermalData` is still imported and used.
3. + 4. `raw` survives on line 13 in both routes and the `data` expression is identical — the deleted `query.raw` was assigned from that same `raw` const, so `raw ? ... : transformThermalData(...)` evaluates the same. On `query.timeout`: `fetchDataCommon(unit, req.query)` is called on line 11 *before* the object was built, and it destructures `timeout` off `req.query` itself (`fetchFromUnits.ts:20`). The deleted object was never passed to anything, so it was not the thing feeding the timeout. Request timeout behaviour is unchanged.
5. `i = tryNumber;` — `i` is declared with `let` inside the async `reduce` callback (`fetchFromUnits.ts:145`), so it is per-iteration and dies with the closure. Nothing outside the callback reads it, and no inner function captures it past the `while` loop. Each iteration re-initialises `i` from `tryNumber` anyway. Dead.
6. `updateWatchdogEngine(timeout = 5000)` — private method, one call site (`panelLoop.ts:208`) with no arguments, never used as a callback reference. The panel `updateWatchdog()` it calls takes no parameters, so the value had nowhere to go.

**note** — `EngineData` in `source/apps/backend/src/types/data.types.ts` now has zero users (word-boundary grep; the `EngineData*` hits elsewhere are `EngineDataRequestQuery` and the panel's unrelated `getEngineData`). Fine to leave, but it is a follow-up.

**note** — no `node_modules` in the worktree, so no build or typecheck was run here either. Small diff, and every removal is grep-confirmed, but CI should be the gate.

**note** — commit hygiene is fine: one commit, subject only, no body, no `Co-Authored-By`. The PR *description* still carries the "Generated with Claude Code" line; harmless, but strip it if you keep descriptions clean too.

**Scope** — the diff is exactly the six items across six files, nothing extra.

PR #59, opened 2026-09-09
chore: move vite and vue-tsc to devDependencies

zbyju:
Moves `vite` and `vue-tsc` from `dependencies` to `devDependencies` in the panel app. Both are build tools — they only run during `dev`, `build` and `typecheck`, never at runtime.

No version numbers changed. `vite` stays `^2.9.5` and `vue-tsc` stays `^0.38.2`. The new entries keep the existing alphabetical order in `devDependencies`.

The lockfile was NOT refreshed. The `pnpm install --lockfile-only` step could not run in this environment, so `pnpm-lock.yaml` still lists both packages under the panel importer's `dependencies` block. Please run `pnpm install --lockfile-only` locally and commit the result before merging.

Verification: read the file, made the edit, checked `git diff` showed only the four moved lines, and validated the file still parses as JSON.


zbyju (commented):
**Verdict: the `package.json` move is correct. The lockfile really is stale and it is still stale on `main` — needs a follow-up commit. (Posting after merge.)**

**[non-blocking — correct]** Both entries moved, versions unchanged (`vite ^2.9.5`, `vue-tsc ^0.38.2`), alphabetical order kept, file still parses as JSON, nothing else touched. Nothing under `source/apps/panel/src` imports either package. The only hits outside `package.json` are `vite.config.ts` (build config) and `env.d.ts` (`vite/client` types) — both build-time, both fine. The move is right.

**[blocking for a follow-up]** The author's claim checks out on both counts:

- The `apps/panel` importer in `source/pnpm-lock.yaml` has separate `dependencies:` / `devDependencies:` blocks, and still lists `vite: 2.9.14_stylus@0.57.0` and `vue-tsc: 0.38.9_typescript@4.7.4` under `dependencies`.
- Package entries carry `dev:` flags. `/vue-tsc/0.38.9_typescript@4.7.4` is `dev: false`, which is now wrong, and so is the `@volar/*` subtree under it.

pnpm does not tolerate the mismatch. I ran the check with the pinned pnpm:

```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile"
because pnpm-lock.yaml is not up-to-date with apps/panel/package.json
```

That is a hard failure, not a warning.

**[note — how bad in practice]** Nothing in this repo installs with a frozen lockfile. There is no `.github/`, and `.travis.yml` is gone as of #61. The deploy path (`source/apps/startup/src/logic/start/{ubuntu,windows}.js`) and the `build` / `start:main` scripts in `source/package.json` all run plain `pnpm install`, which just repairs the lockfile in place. So no device deploy is at risk today. The exposure is that pnpm sets `--frozen-lockfile` by default when `CI=true`, so the first CI job anyone adds will fail on this.

**[refresh command]** Do not use a locally installed pnpm 10 — it rewrites the file to `lockfileVersion: 9.0`, thousands of lines. `source/package.json` pins `pnpm@7.5.0`, so use that:

```
cd source && npx --yes pnpm@7.5.0 install --lockfile-only
```

Or, honouring the `packageManager` pin: `cd source && corepack enable && corepack pnpm install --lockfile-only`.

I ran the `npx` form against a copy of the tree. It stays at `lockfileVersion: 5.4` and produces about 200 changed lines: the two importer entries moving, plus `dev: true` flips on the `vue-tsc` / `@volar/*` / `vite` transitive tree. Zero version and zero integrity changes. Small and reviewable.

**[note — hygiene]** Commit is clean: `chore: move vite and vue-tsc to devDependencies`, subject line only, no body, no trailers. The PR description carries a "Generated with Claude Code" line; keep that out of commit messages (it is).

PR #60, opened 2026-09-09
chore: stop tracking compiled configer JavaScript

zbyju:
Deletes seven compiled `.js` files that sat next to their `.ts` sources in `source/apps/configer/src/`, and adds `apps/configer/src/**/*.js` to `source/.gitignore` so they cannot come back.

They were stale build output, last touched in January 2023, and they made two copies of the same code visible in the repo. That is a trap: a reader or a tool can pick the wrong one.

Nothing loads them. `dev` runs `nodemon --esm ./src/index.ts`, so ts-node maps the `.js` import specifiers back to the `.ts` files. `build` is `tsc` with `outDir: ./dist` and `allowJs` off, so it never reads them and never writes into `src/`. `start` is `pm2 start ./dist/index.js`. I checked all three.

Risk: low. No file in the repo points at `source/apps/configer/src/*.js`.

Verification: read `source/apps/configer/package.json` (dev/build/start scripts), `tsconfig.json` (`outDir: ./dist`, `include: ["src"]`, `allowJs` commented out), and every import in the configer TS sources; grepped the whole repo for references to configer `src/*.js` and found none; the only `node src/index.js` calls are in the `startup` app, which has no TypeScript at all; confirmed the new ignore rule catches a fresh `.js` in that folder with `git check-ignore`.


zbyju (commented):
**Verdict: sound, no blocking findings.** The three entry points are safe, and the `dev` path is not just safe — it was actively broken by these files, so this is a fix.

**[non-blocking] `dev` was loading the stale `.js`, not the `.ts`.** Worth correcting the PR description, which says ts-node "maps the `.js` import specifiers back to the `.ts` files". It only does that when the `.js` is absent. ts-node 10.9.1 maps a `.js` specifier through `resolveReplacementExtensions` (`dist-raw/node-internal-modules-esm-resolve.js`), which walks `replacementsForJs` (`dist/file-extensions.js`) in `preferTsExts` order and takes the first file that exists. `preferTsExts` is not set here, so that order is `['.js', '.ts']` — **`.js` wins when both are on disk**. I reproduced it: a stale `dep.js` beside `dep.ts`, imported as `./dep.js`, loaded the `.js`. So before this PR every relative import in `src/index.ts` resolved to January-2023 output, and editing those `.ts` files did nothing in dev. After the deletions the specifiers resolve to `.ts`. Deleting them is strictly correct.

**[note] `build` and `start` untouched.** `allowJs` is commented out in `tsconfig.json`, `include: ["src"]`, `outDir: ./dist`. Reproduced with tsc 4.7.4: a `.js` in `src` is not an compile input and is not emitted; output is `dist/index.js`, which is what `pm2 start ./dist/index.js` expects. Nothing else in the repo points at `configer/src/*.js` — the only `src/index.js` references are in `startup`, which has no TypeScript.

**[note] Ignore rule is correctly anchored and not too wide.** `git check-ignore -v` catches `source/apps/configer/src/newthing.js` and `source/apps/configer/src/deep/er/nested.js`, and leaves alone `configer/scripts/hook.js`, `configer/src/notes.md`, `apps/backend/src/foo.js` and `apps/panel/src/foo.js`. That last part matters: `apps/startup` is plain JS with `src/index.js` as its entry point, and it stays tracked. A hand-written `.js` in `configer/src` could not work anyway with `allowJs` off, so the rule is right-sized. The one cost is that if someone ever adds one, `git add` will fail quietly — acceptable here.

**[note] Deletions are exactly right.** Seven files, a clean 1:1 with the seven `.ts` sources, no orphan `.js` left and none missed. Spot-checked `index.js` against `index.ts` (types stripped, same shape) and `versions.types.js` is `export {};`, which is what tsc emits for a types-only ESM file. Nothing hand-edited got caught.

**[note] Comment earns its place.** "configer is TypeScript only; any .js in its src is stale compiler output" gives the why, not a restatement of the glob. Keep it.

**[note] Commit is clean** — one commit, subject only, no body, no trailers. But the PR *description* still carries the "Generated with Claude Code" line. Cosmetic, and the commit itself is fine.

**[note] Pre-existing, not this PR:** `nodemon` is not declared in any `package.json`, and ts-node 10.9.1 `--esm` fails on Node >= 20 with `ERR_UNKNOWN_FILE_EXTENSION` with or without these files. So `pnpm dev:configer` cannot be smoke-tested on a current Node either way. Also `source/apps/configer/.env` is tracked — out of scope here.

zbyju:
Correction for the record, from the review on this PR: the rationale in the description was backwards in one respect.

Those `.js` files were **not** inert — `pnpm dev:configer` was loading them. ts-node 10.9.1 resolves a `.js` import specifier by walking `['.js', '.ts']` and taking the first extension present on disk. With `preferTsExts` unset, `.js` wins whenever both exist. This was reproduced in a scratch project matching configer's exact setup.

So dev has been running January-2023 compiler output, and edits to `factory.ts`, `main.ts`, `configRoute.ts` had no effect there. `build` and `start` were always safe — they use `dist/`.

The description's claim that ts-node maps `.js` specifiers back to the `.ts` files is true only *after* these files are gone. This PR was a bug fix, not just hygiene.

PR #61, opened 2026-09-09
chore: remove dead Travis config and committed Coveralls token

zbyju:
Deletes `source/apps/backend/coveralls.yml`, which held a live Coveralls `repo_token` in plain text.

Deletes `source/apps/backend/.travis.yml`. It is dead: it runs `pnpm test` and `pnpm coveralls`, and neither script exists at the workspace root. The project no longer uses Travis.

Drops the `coveralls` script and the `coveralls` devDependency from the backend `package.json`, and removes the two dead Travis/Coveralls badges from the backend README. The `test` script and jest are untouched.

**Please rotate the Coveralls repo token.** Deleting the file does not remove it from git history, so the old token stays readable to anyone who can read this repo.

Risk: low. Nothing in the build, dev, or test path referenced these files. `pnpm-lock.yaml` was left alone on purpose: the local pnpm (v10) rewrites the committed v5.4 lockfile to v9.0, an ~8600-line change well outside the scope of this PR. The stale `coveralls` entries in the lockfile are harmless and should be cleared by whoever next bumps the lockfile format.

Verification: grepped the repo for `coveralls` and `travis` (no hits left outside `pnpm-lock.yaml`); confirmed the edited `package.json` still parses; reviewed the full staged diff.


zbyju (commented):
Checks out. Merge it — but rotate the Coveralls token first, not after.

**Travis is dead — confirmed twice over.** [non-blocking / verified]
- No CI in the repo at all: no `.github/`, no GitLab/Circle/Jenkins/Azure config. The only YAML files are the two lockfiles and `pnpm-workspace.yaml`.
- Travis only reads `.travis.yml` from the **repo root**. This one sat at `source/apps/backend/.travis.yml`, so it would never have been picked up even if the project were still connected.
- Its `script:` ran `pnpm test` and `pnpm coveralls`; `source/package.json` has neither (only `dev*`, `build`, `start:*`, `stop`, `echo`).

**Nothing broke from dropping the `coveralls` script.** [verified]
`test: "jest"` and `jest.config.json` are untouched, the edited `package.json` parses, and `grep -rn -i coveralls` over the tree returns zero hits outside `pnpm-lock.yaml`.

**note:** `jest.config.json` still sets `collectCoverage: true`. That was there to produce `coverage/lcov.info` for Coveralls. Nothing consumes it now, so every `pnpm test` still writes a gitignored `coverage/` dir for no reader. Not broken — just the other half of this cleanup. Fine either way.

**README badge removal: keep it in this PR.** [opinion]
The badges pointed at `zbyju/babybox-panel-backend` (the old standalone repo) and at exactly the two services being deleted here. Leaving them means two broken badges advertising CI that does not exist. It is one line, same subject, and splitting it would be busywork. Encoding survived: still UTF-8 with BOM (`ef bb bf`), all 26 lines CRLF, line 1 ends `0d 0a`, and `--stat` shows 1 insertion / 1 deletion. No mangling.

**The secret.** [non-blocking, but do it before merge]
The PR body does tell the maintainer to rotate, in bold. Verified the token value appears nowhere in the PR title, body, or commit message, and nowhere in the tree at head. Deleting the file does not touch history, so treat the token as compromised and rotate it in Coveralls before this lands — otherwise the merge just makes it harder to notice.

**Lockfile: note, not a blocker.** [note]
`source/pnpm-lock.yaml` still has 3 `coveralls` entries. Nothing in the repo uses `--frozen-lockfile` or `npm ci` — every install path is a plain `pnpm install` (root `build` / `start:main`, `scripts/ubuntu/install*.sh`, `scripts/windows/install.bat`, `src/logic/start/{ubuntu,windows}.js`), which reconciles a removed devDep on its own. Stale and inert. One caveat for later: pnpm turns on frozen-lockfile by default when `CI=true`, so if CI is ever added, that install will fail until the lockfile is refreshed. Deferring is reasonable given the v5.4 to v9.0 rewrite.

**Commit hygiene: clean.** One commit, conventional subject, empty body, no trailers, no "Generated with" line.

**Unrelated note:** `source/apps/backend/.env` is tracked. I checked it — only `NODE_ENV`, `PORT`, `API_PREFIX`, no secrets. Not this PR's job, but a tracked `.env` is how the next real secret gets committed.

PR #62, opened 2026-09-09
fix: clear camera interval on unmount

zbyju:
**Stacked on #58 — merge that first. Base will need retargeting to `main` after #58 merges.**

`useCamera` called `setInterval` and never kept the id, so every mount of a camera component left a permanent timer writing to a ref nobody reads.

On a panel that runs unattended for weeks, those timers accumulate for the whole uptime — the fix keeps the id and clears it in `onUnmounted`.

Also removed the dead commented-out `if (config.cameraType === "dahua") { ... } else {` block and its trailing `// }`, consistent with the `stringToCameraType` cleanup already on `main`.

Only call site is `source/apps/panel/src/components/panel/elements/CameraView.vue:70`, at the top level of `<script setup>`, so `onUnmounted` registers correctly and no cleanup function needs returning.

URL shape and update cadence are unchanged.

Verification: no typecheck ran — `source/apps/panel/node_modules` is absent and `pnpm install` was skipped as too slow. Instead grepped `useCamera` across `source/apps/panel/src` (one import, one call), read the call site, and re-read the diff.

Separate issue, NOT fixed here: `config.updateDelay` goes straight into `setInterval`. It is typed `number` and `base.json` sets `1000`, but the runtime guard `isInstanceOfCameraConfig` only checks the key exists, so a hand-edited `main.json` with `0` or `null` would make the interval fire every event-loop tick.


zbyju (commented):
Leak is genuinely fixed. No blocking findings.

**Verified independently**

- `non-blocking` (confirmed, no change needed) Timer id is captured and cleared. One `setInterval` per `useCamera` call, one `onUnmounted` per call, no path that starts a second interval without clearing the first — the composable is never re-run for a live component.
- `non-blocking` (confirmed) Exactly one call site: `CameraView.vue:70`, a synchronous top-level statement in `<script setup>`. No `await` before it, setup is not async, so `onUnmounted` registers against the right instance. No `<KeepAlive>` anywhere in `source/apps/panel/src`, and `CameraWrapper.vue` renders `CameraView` unconditionally. The leak was real: `CameraView` sits under both `MainView` and `SettingsView`, so every route change between Main / Settings / Data started another permanent timer.
- `non-blocking` (confirmed) Behaviour unchanged. Same template literal, same `getURLPostfix(cameraType)`, same `config.updateDelay`. The removed `if/else` was fully commented out, so it never nested the live body — deleting it changes nothing at runtime.

**Findings**

- `non-blocking` Reword the new JSDoc line. As written it half-states what a Vue developer already knows ("must be called from a component's setup"). The part worth a reader's time is what the signature hides: `(config, onUpdate?) => Ref<string>` gives no hint that a timer is started, and the failure outside setup is silent in production. Suggest replacing it with something like: `Starts a refresh timer that is only stopped on the calling component's unmount.` One line, same length, carries the non-obvious half.
- `note` Your `updateDelay` reading is correct. `isInstanceOfCameraConfig` in `source/apps/panel/src/utils/panel/instanceCheck.ts:49` only tests key presence, so `0` or `null` in a hand-edited `main.json` passes the guard and reaches `setInterval`. Strictly it is not an unbounded busy loop — the HTML spec clamps nested timers to ~4 ms — but ~250 URL rebuilds per second on panel hardware is bad enough. Agree with leaving it out: every section of that guard validates by key presence only, so a real fix means value checks across the whole file, not a patch inside this composable.
- `note` Out of scope, worth knowing: `useCamera` runs even when `cameraType === "vivotek"`, where `CameraView` renders `VivotekCameraView` and `url` is never read. Pre-existing, and this PR at least bounds that timer to the component's lifetime.

**Scope and hygiene**

- Diff is one file, 6 added / 6 removed, nothing beyond the two stated items plus the doc line.
- Commit message is a clean conventional subject line, no body, no trailers. The PR description still carries the "Generated with Claude Code" line — same convention, worth dropping.

zbyju:
Reworded the JSDoc line per the review: it now says `Starts a refresh timer that is only stopped on the calling component's unmount.` instead of stating the composable must be called from setup.

Nothing else changed - the interval logic, URL building and the dead-code removal are untouched.

zbyju:
Rebased onto main so the diff shows only this change. No file content changed.

zbyju:
## Build verification (no CI in this repo, so this was run by hand)

Compiled this branch and `main` with the same toolchain and diffed the results.

| | Backend `tsc` | Panel typecheck |
|---|---|---|
| `main` | pass (exit 0) | 17 errors |
| this branch | pass (exit 0) | 17 errors |

**Errors on this branch but not on `main`: none.** The panel error lists are identical (file + code, line numbers normalised). Changed files confirmed present in the compile set via `--listFiles`.

Two things found while doing this, both pre-existing and unrelated to this PR:

1. **`pnpm build`'s typecheck step is a no-op.** `panel/tsconfig.json` is `{"files": [], "references": [...]}`, so `vue-tsc --noEmit` without `--build` checks the root project, which has no files. `npx vue-tsc --noEmit --listFiles | grep -c '/src/'` returns 0. The real check is `npx vue-tsc --noEmit -p tsconfig.app.json --composite false`.
2. **The panel has 17 pre-existing type errors on `main`.** ~13 trace to one environment problem — chai's `Object.prototype` augmentation leaking from a parent `node_modules` adds `should` to `Moment`, poisoning `EngineUnit`, `ThermalUnit` and `LogEntry`. The code-level ones are `useCamera.ts` TS1371, `table.ts` TS1208, and two TS2322s in `SettingsFormTable.vue`.

PR #63, opened 2026-09-09
refactor: extract shared service URLs and fetch timeout defaults

zbyju:
**Stacked on #56 — merge that first. Base will need retargeting to `main` after #56 merges.**

Pure refactor: removes duplicated hardcoded URLs and timeout defaults.

- **Backend configer URL** — new `CONFIGER_API_URL` in `source/apps/backend/src/fetch/constants.ts`, used by `fetchConfig.ts`.
- **Backend fetch timeout** — `parseInt(process.env.DEFAULT_FETCH_TIMEOUT) || 5000` appeared three times in `fetchFromUnits.ts`. Replaced with `defaultFetchTimeout()` in the same constants file. It stays a function, not a const, because `dotenv.config()` runs after module load. The `NaN || 5000` fallback is untouched, so a missing or non-numeric env value still gives 5000.
- **Panel configer URL** — new `CONFIGER_API_URL` in `source/apps/panel/src/api/base.ts`, used by `getConfig()` and `getVersions()` in `panelLoop.ts`. Separate package, so it cannot share the backend constant.
- **Panel API base block** — new `backendApi()` in `source/apps/panel/src/api/base.ts` returns `{ baseUrl, timeout, isConfigured }`. It replaces the repeated four-line block in all eight functions in `api/units.ts` and in `api/restart.ts`. It is a function, not a module-level const, because `useConfigStore()` needs Pinia active.

No behaviour change. Kept deliberately:

- `sendSettings` still uses its own `timeout: 60000`, not `requestTimeout`. It is not routed through the helper default. Added one short comment saying why.
- `updateWatchdog` keeps the `axios.put(url, null, { timeout })` shape from #56.
- `restart.ts` keeps its guard. The `port === undefined || url === undefined` check moved into the helper as `isConfigured`, same expression, same short-circuit before any request.

Verification: no typecheck ran — `node_modules` is not installed in this checkout and `pnpm install` was out of scope. Confirmed by reading instead: grepped every call site of the changed exports (`App.vue`, `panelLoop.ts`, `QuickActions.vue`, `SettingsForm.vue`, backend routes) and no exported name or signature changed; grepped for remaining `5001` and `localhost:${` hardcodes, only the two new constants remain; no test file references the changed modules. The pre-existing TS error in `source/apps/backend/src/types/request.types.ts` was not touched.


zbyju (commented):
Checked every call site against `fix/watchdog-axios-timeout-config`. The "no behaviour change" claim holds — no blocking findings.

**URLs — provably identical**

| before | after |
|---|---|
| `http://localhost:5001/api/v1/config/main` (backend `fetchConfig`) | `${CONFIGER_API_URL}/main` → same |
| `.../config/main`, `.../config/version` (panel `getConfig`/`getVersions`) | same |
| 8x `http://localhost:${port}${url}/<path>` in `units.ts` | `${baseUrl}/<path>` → same |
| `.../restart/refresh` | same |

No doubled or dropped `/`: both constants stop before the slash and every call site supplies its own leading `/`. `git grep 5001` leaves only the two new constants (plus configer's own `.env` and `base.json`); `git grep 'localhost:${'` leaves only `base.ts:17`.

**Timeouts — identical.** `defaultFetchTimeout()` is byte-for-byte the old expression, so `NaN || 5000` still gives 5000 for a missing or non-numeric value. Panel `requestTimeout || 5000` survived as `DEFAULT_REQUEST_TIMEOUT`. `sendSettings` still passes only `{ timeout: 60000 }` and never touches the helper's timeout. `updateWatchdog` still reads `axios.put(url, null, { timeout })`.

**Guard.** `!(port !== undefined && url !== undefined)` is exactly the old `port === undefined || url === undefined`. It returns before `axios.get`, so nothing is sent while unconfigured.

**Pinia.** `useConfigStore()` sits inside `backendApi()`, not at module scope. No new top-level store call anywhere.

**Commit.** Subject line only, no body, no trailers. Correct.

---

- **note — the dotenv argument is right, but not for the reason the description gives.** `dotenv.config()` is at `index.ts:23` inside `main()`, and PR #57 does not move it (#57 only touches `RESTART_ENABLED` and the `RESTART_INTEVAL` typo). Even if it were moved to module scope in `index.ts`, the compiled CommonJS `require`s still run `constants.ts` first, so a module-level const would read `undefined`. Keep the function. The comment is not wrong; tighten "runs later, during startup" only if you want it exact.
- **note — `baseUrl` is built before `isConfigured` is read.** Harmless today because nothing sends, but the string is `http://localhost:undefinedundefined` when unconfigured. If anyone ever logs it, that is a confusing line in a hospital log. Optional: return `baseUrl: undefined` when not configured.
- **note — both comments earn their place.** The `sendSettings` one matches reality (`updateSettings` uses `tryNumber = 10`), and the `backendApi()` one flags a real Pinia trap on a shared helper. Keep both.
- **note — placement is fine.** `fetch/constants.ts` sits with `fetch.ts` and `fetchConfig.ts`; `api/base.ts` sits with `units.ts` and `restart.ts`. Two constants for two packages is right — panel and backend share no module. Minor: `constants.ts` also exports a function, so the name reads slightly off; not worth a re-push.
- **note — left alone, correctly.** `updateSettings`, `isReady`, `updateSetting` (backend) and `getData` (panel) keep `timeout = 5000` as parameter defaults. Different thing from the env default; leaving them is right.
- **note — PR body carries a "Generated with Claude Code" line.** The commit is clean, so nothing to fix in git history; strip it from the description if the repo wants that.

zbyju:
Rebased onto main so the diff shows only this change. No file content changed.

zbyju:
## Build verification (no CI in this repo, so this was run by hand)

Compiled this branch and `main` with the same toolchain and diffed the results.

| | Backend `tsc` | Panel typecheck |
|---|---|---|
| `main` | pass (exit 0) | 17 errors |
| this branch | pass (exit 0) | 17 errors |

**Errors on this branch but not on `main`: none.** The panel error lists are identical (file + code, line numbers normalised). Changed files confirmed present in the compile set via `--listFiles`.

Two things found while doing this, both pre-existing and unrelated to this PR:

1. **`pnpm build`'s typecheck step is a no-op.** `panel/tsconfig.json` is `{"files": [], "references": [...]}`, so `vue-tsc --noEmit` without `--build` checks the root project, which has no files. `npx vue-tsc --noEmit --listFiles | grep -c '/src/'` returns 0. The real check is `npx vue-tsc --noEmit -p tsconfig.app.json --composite false`.
2. **The panel has 17 pre-existing type errors on `main`.** ~13 trace to one environment problem — chai's `Object.prototype` augmentation leaking from a parent `node_modules` adds `should` to `Moment`, poisoning `EngineUnit`, `ThermalUnit` and `LogEntry`. The code-level ones are `useCamera.ts` TS1371, `table.ts` TS1208, and two TS2322s in `SettingsFormTable.vue`.

PR #64, opened 2026-09-09
chore: refresh pnpm lockfile after dependency moves

zbyju:
## What went stale

Two merged PRs changed dependency declarations but did not refresh `source/pnpm-lock.yaml`:

- #59 moved `vite` and `vue-tsc` from `dependencies` to `devDependencies` in `source/apps/panel/package.json`.
- #61 removed the `coveralls` devDependency from `source/apps/backend/package.json`.

The lockfile on `main` no longer matches those files. Nothing installs with `--frozen-lockfile` today, so no deploy is broken. But pnpm turns `--frozen-lockfile` on by default when `CI=true`, so the first CI job anyone adds would fail with `ERR_PNPM_OUTDATED_LOCKFILE`.

## What this changes

Only `source/pnpm-lock.yaml`. No `package.json` was touched.

Command used:

```
npx --yes pnpm@7.5.0 --dir source install --lockfile-only
```

pnpm 7.5.0 matches the `packageManager` pin in `source/package.json`. The lockfile stays at `lockfileVersion: 5.4`. Running the machine's pnpm 10 would rewrite the whole file to `lockfileVersion: 9.0` (~8600 changed lines), so that was avoided.

Diff: 54 insertions, 89 deletions in one file. Every change is one of:

- `vite` and `vue-tsc` moving from the panel importer's `dependencies:` block to its `devDependencies:` block
- `dev:` flags flipping to `true` for packages now reachable only through dev deps
- `coveralls` and its now-unused transitive deps (`request`, `form-data@2.3.3`, `http-signature@1.2.0`, `jsprim@1.4.2`, `lcov-parse`, `log-driver`, `tunnel-agent`) being removed

**No package version changed and no `integrity` hash changed.** The diff adds zero `integrity` lines; the only removed ones belong to whole package blocks that were dropped with `coveralls`.

## Verification

**Correction:** an earlier version of this description cited `pnpm@7.5.0 install --frozen-lockfile --lockfile-only --offline` exiting 0 as proof. That claim was wrong. Review on this PR showed pnpm 7.5.0 silently ignores `--frozen-lockfile` when `--lockfile-only` is also passed — it exits 0 even against a stale lockfile, and rewrites the file. That command proved nothing.

What actually holds:

- Regenerating from this branch with `npx --yes pnpm@7.5.0 --dir source install --lockfile-only --offline` produces a **byte-identical** file (empty `git diff`). The lockfile is at a fixed point.
- Starting from `main`'s stale lockfile, pnpm rewrites it into exactly this PR's content. So the `dev:` flips are pnpm's own computation, not hand edits.
- Both sides parsed into `name/version -> integrity` maps and compared as sets: 0 packages added, 0 integrity changes on surviving packages, 8 removed. `form-data@4.0.0`, `http-signature@1.3.6` and `jsprim@2.0.2` survive and are still referenced; no dangling references.
- `head -5 source/pnpm-lock.yaml` still shows `lockfileVersion: 5.4`; `grep -c coveralls source/pnpm-lock.yaml` returns 0.


zbyju (commented):
**Verdict: clean. No blocking findings.** I independently confirmed zero version changes and zero integrity changes on every surviving package.

### What I checked

**Format** — `head -5 source/pnpm-lock.yaml` reads `lockfileVersion: 5.4`. No rewrite to 9.0.

**No silent upgrades** — I parsed both sides of the diff into `name/version -> integrity` maps and compared the sets rather than reading the hunks by eye:

- packages added: **0**
- integrity changed on a surviving package: **0**
- packages removed: **8**, exactly the set you list — `coveralls@3.1.1`, `request@2.88.2`, `form-data@2.3.3`, `http-signature@1.2.0`, `jsprim@1.4.2`, `lcov-parse@1.0.0`, `log-driver@1.2.7`, `tunnel-agent@0.6.0`

The three duplicated names lost only the coveralls-side copy; the newer one survives and is still referenced (`form-data@4.0.0`, `http-signature@1.3.6`, `jsprim@2.0.2`). `js-yaml@3.14.1` and `minimist@1.2.6` were coveralls deps but stayed, correctly, since other packages still need them.

**Reachability** — every reference to a removed package in the old lockfile sat inside another removed block or the backend importer. Nothing dangles: grepping the new lockfile for those exact versions returns nothing. `jest`, `newman` and `winston` are untouched — `newman` goes through `postman-request`, not `request`.

**Every hunk explainable** — strongest evidence here is reproduction, not reading. I checked out the branch and ran `pnpm@7.5.0 --dir source install --lockfile-only --offline`. The regenerated file is **byte-identical** to the committed one (`git status` clean). So the ~50 `dev:` flips are pnpm's own computation from the current `package.json` files, not hand edits.

**Does it fix the problem** — yes. I then overwrote the branch's lockfile with `main`'s stale one and re-ran the same command: pnpm rewrote it back into exactly this PR's content. That is direct proof the committed lockfile is what pnpm 7.5.0 derives from the current manifests.

**Scope** — one file, `source/pnpm-lock.yaml`. No `package.json` touched.

**Commit hygiene** — one commit, subject line only, no body, no `Co-Authored-By`. Matches the convention.

### Findings

- **non-blocking** — the verification command in the PR description does not prove what it claims. `--frozen-lockfile` is silently ignored when `--lockfile-only` is also passed on pnpm 7.5.0. I ran `install --frozen-lockfile --lockfile-only --offline` against `main`'s **stale** lockfile: exit 0, and it rewrote the file instead of failing. So "the retest exits 0" is true on the broken lockfile too. Swap the claim for the reproduction check (regenerate, then `git diff` is empty), which does hold.

- **note** — `ERR_PNPM_META_FETCH_FAIL` claim confirmed. `graceful-fs@4.2.10` and `uglify-js@3.16.3` are byte-identical on both sides and appear nowhere in the diff. `fsevents@2.3.2` gains only `dev: true`; version and integrity unchanged. Those packages did resolve from the existing lockfile.

- **note** — the `dev:` flips have no effect on the devices today. Nothing in the repo passes `--prod` or `--production`; every install path (`start:main`, `build`, the startup scripts) is a plain `pnpm install`, so dev deps are installed anyway. `vite` moving to devDependencies does not break `pnpm run build`.

- **note, pre-existing, out of scope** — `source/apps/panel/pnpm-lock.yaml` is tracked. A nested lockfile inside a workspace package is ignored by the root install, so it drifts silently and will confuse the next person who greps for a version. Worth deleting in a separate PR.

- **note** — there are no GitHub Actions workflows in the repo, so the `CI=true` failure this prevents is still hypothetical. That matches what the description says; flagging only so the urgency is read correctly.

PR #65, opened 2026-09-09
fix: load .env before reading the restart env vars

zbyju:
Follow-up to #57.

#57 fixed the `RESTART_ENABLED` predicate, but the switch still could not be set through `.env`. `export const modules = modulesObject();` runs at module scope in `src/index.ts`, while `dotenv.config()` sat inside `main()` — so `RESTART_ENABLED`, `RESTART_INTERVAL` and `RESTART_ERROR_THRESHOLD` were read before `.env` was parsed. Since `.env` is the only place this backend is configured, the kill switch could never actually be turned off. First commit moves `dotenv.config()` above the `modulesObject()` call and drops it from `main()`.

Second commit widens the accepted "off" values: `RESTART_ENABLED` is now trimmed and lowercased and matched against `false`, `0`, `no`, `off`. Unset still means enabled.

Verification: `tsc -p tsconfig.json` builds with zero errors in `src` (only pre-existing parse noise from a `bun-types` package in a parent `node_modules`); ran the built `dist/index.js` with a temporary `.env` of `RESTART_ENABLED=OFF` and `RESTART_INTERVAL=99000` and a temporary probe, which printed `restartEnabled: false, interval: '99000'`; re-ran with `RESTART_ENABLED` unset and got `restartEnabled: true`; probe, temporary `.env` and `dist` removed afterwards. Diff touches only `src/index.ts` and `src/modules/init.ts`.


zbyju (commented):
**Verdict: correct, and I confirmed it at runtime — no blocking findings. Two things worth deciding before merge.**

**Verified empirically, not just by reading.** Built with the repo's own tsc 4.7.4 (clean), then ran `dist/index.js` from a scratch cwd with a temporary `.env`, with `setInterval` and `child_process.exec` stubbed so nothing could reboot. `RESTART_ENABLED=OFF, RESTART_INTERVAL=99000` gave timer delay `99000` and `onIncomingRequest` not called; unset gave it called. I then flipped the two lines back in `dist/index.js` only and re-ran: delay fell back to `20000` and the switch was ignored. So the reorder is load-bearing, and the old code really was broken. Probe, temp `.env`, `dist` and `node_modules` removed; `git status` clean; tracked `source/apps/backend/.env` never touched.

- **[non-blocking] Ordering holds in the emitted CJS.** All `require` calls hoist above both statements, so `dotenv.config()` now precedes `exports.modules = (0, init_1.modulesObject)()`. Nothing reads the RESTART_* vars at import time: `init.ts` reads inside `isRestartEnabled()`, `restart.ts` reads `RESTART_INTERVAL` / `RESTART_ERROR_THRESHOLD` inside the `restartRepository()` body. Both only run when `modulesObject()` is called.

- **[non-blocking] The `restart.ts` → `".."` cycle is unaffected.** tsc emits `__1.config`, a live property read, not a destructured snapshot, and `config` is only dereferenced inside `startRestart` / `stopRestart`, which run from the timer. Moving a statement within `index.ts`'s body does not change which modules load or in what order — `restart.js` was already required before `dotenv.config()` under both orderings.

- **[non-blocking] No other env read is stranded.** `PORT`, `NODE_ENV`, `API_PREFIX` are all inside `main()`, which runs after. The three `DEFAULT_FETCH_TIMEOUT` reads in `fetch/fetchFromUnits.ts` are all inside function bodies (request time). `dotenv.config()` does not override vars already in `process.env`, so running it earlier cannot regress anything. Exactly one `dotenv.config()` in `src` — no leftover in `main()`.

- **[non-blocking] Predicate gap.** Unset means enabled, and `?? ""` handles `undefined`. I measured the accepted set: `off`, `OFF`, `No`, `0`, `FALSE` disable; `disabled`, `n`, and an empty `RESTART_ENABLED=` still mean **enabled**. Empty is the one I'd fix — an operator who comments out the value rather than the line gets a rebooting machine. Better than widening the list further: log the resolved state once at boot, so the value is visible instead of guessed. The implementation itself is the right size for this codebase.

- **[non-blocking] The switch does stop a reboot today, but by accident, not by design.** `restartRepository()` is still called unconditionally and registers its `setInterval` either way; the flag only gates `onIncomingData()`. The single path to `onIncomingRequest` is `GET .../restart/refresh` → `modules.onIncomingData()`, so with the flag off `lastRequest` stays `null` and the timer's `if (lastRequest === null) return;` short-circuits before `errorStreak` ever moves. Safety rests on that guard, not on the flag. I agree with deferring the restructure — this diff is the part that actually unblocks the switch and it stays reviewable — but the follow-up should skip building the repository at all when disabled, so the guarantee is structural. Worth opening that issue now rather than after merge.

- **[note] `source/apps/backend/.env` (tracked) has no RESTART_* keys.** The switch this PR fixes is invisible to whoever edits that file on site. A commented `# RESTART_ENABLED=false` line would make it findable.

- **[note] Scope and commits are clean.** Exactly the two files, nothing from other merged PRs. Both commits are subject-only conventional, no body, no trailers. The PR description carries a "Generated with Claude Code" line; the commits don't.

- **[note] Comments.** The `dotenv` one earns its place — it stops a future reader tidying the call back into `main()`. I'd make the consequence explicit: "otherwise the switch is read before `.env` is parsed". The `init.ts` block is a touch long but states a real why; fine as is.

PR #66, opened 2026-09-09
fix: sync pnpm-lock.yaml with the package.json files

zbyju:
`pnpm install --frozen-lockfile` fails on `main` today. This commit makes the lockfile match the `package.json` files again. **No package was upgraded** — the diff removes entries, it adds no resolution.

## The problem

Two merged PRs changed a `package.json` without regenerating the lockfile:

| PR | Change | Lockfile still says |
|---|---|---|
| #61 | dropped `coveralls` from `apps/backend` | `coveralls: ^3.1.1` and `coveralls: 3.1.1` |
| #59 | moved `vite` / `vue-tsc` to `apps/panel` devDependencies | both under `dependencies` |

On `main` as of `2c4cff2`:

```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile"
because pnpm-lock.yaml is not up-to-date with apps/backend/package.json
```

Nothing caught this, because no job runs `--frozen-lockfile`.

## The fix

Regenerated with pnpm 7.5.0 on Node 18.12.1 — the versions the panel computers run. 54 added, 89 removed, all in `source/pnpm-lock.yaml`:

- `coveralls` gone, plus the transitive deps nothing else needs: `request`, `form-data`, `http-signature`, `jsprim`, `lcov-parse`, `log-driver`, `tunnel-agent`, and their subtree
- `vite` and `vue-tsc` moved to panel's dev section, which flips their `dev:` flags

The diff contains **no `+  /package/version:` line**. Nothing resolves to a new version, so no box installs anything different.

## Testing

```
pnpm install --frozen-lockfile   →  exit 0, "Lockfile is up-to-date"
```

Before the commit the same command exits non-zero with `ERR_PNPM_OUTDATED_LOCKFILE`.

Note that `pnpm install --lockfile-only --frozen-lockfile` silently ignores `--frozen-lockfile` and passes on broken input — it is not a valid check for this.

## Interaction with #50

#50 also rewrites `pnpm-lock.yaml`, so whichever lands second will conflict. The resolution is the same regenerate-and-verify done here. #50 is currently blocked on an open design question, which is why this is split out rather than left to ride along.

## Worth a follow-up

Nothing verifies the lockfile in CI. A single `pnpm install --frozen-lockfile` step would have caught both #59 and #61 before merge, and it matters more now: #50 makes `startup.sh` exit non-zero when the frozen install fails, so a drifted lockfile on `main` would stop panel computers from booting.


zbyju:
Superseded by #64, which landed the same change — 54 added, 89 removed in `source/pnpm-lock.yaml`. The two lockfiles are byte-identical, and `pnpm install --frozen-lockfile` now exits 0 on `main` at `bfae1ac`.

Closing. One thing this PR raised that #64 did not cover: nothing runs `--frozen-lockfile` in CI, so the same drift can recur on the next `package.json` edit. That matters more once #50 lands, because `startup.sh` will then exit non-zero on a failed frozen install instead of booting the panel.

PR #67, opened 2026-09-09
perf: pass units config into getNewState instead of resolving the store per tick

zbyju:
## What / Why

`getNewState()` called `useConfigStore()` and `storeToRefs(configStore)` on every invocation, but only needs two numbers: `units.warningThreshold` and `units.errorThreshold`. It runs from `AppManager.updateState()` on every poll tick (~2s), forever.

The thresholds are now passed in as a `unitsConfig: UnitsConfig` parameter. `AppManager` already holds `this.unitsConfig` as a ref from `storeToRefs` in its constructor, so it just forwards `this.unitsConfig.value`. `getNewState` becomes a pure function of its inputs and no longer depends on Pinia.

## Performance impact

Estimated — not measured on device.

- Removes 1 `useConfigStore()` lookup + 1 `storeToRefs()` call per tick. `storeToRefs` walks every property of the store and allocates a fresh `ObjectRefImpl` (or `toRef`) object per state/getter property, so the saved allocation count equals the config store's property count, not 1.
- At a 2s tick that is ~43,200 calls/day, ~15.8M/year on a box that never restarts. Each one produced short-lived garbage that the GC then had to sweep.
- Bundle: net -1 import (`storeToRefs`, `useConfigStore`) from `state.ts`. Both are still used elsewhere in the panel, so the tree-shaken bundle change is roughly zero bytes. The win is runtime allocations, not size.
- Latency per tick: sub-millisecond either way. This is about steady-state GC pressure over months of uptime, not tick latency.

## Risk

Low.

- Behaviour is byte-identical: the `|| 5` and `|| 25` fallbacks are preserved exactly, so `0` and `undefined` still fall back the same way.
- The thresholds are NOT snapshotted. `this.unitsConfig` is a `Ref` from `storeToRefs`, and `updateState()` reads `.value` fresh on each tick, so a live config change from the settings screen is picked up on the next tick — same as before.
- `getNewState` has exactly one caller (`panelLoop.ts:103`) and no test imports it (checked all `*.spec.ts` / `*.test.ts` in the repo), so the signature change has no other blast radius.

## Verification

- Read every caller: `grep` across the repo found one call site and one import, both in `source/apps/panel/src/logic/panel/panelLoop.ts`. No test file references `getNewState`.
- Confirmed `UnitsConfig` in `source/apps/panel/src/types/panel/config.types.ts` declares `warningThreshold: number` and `errorThreshold: number`, so the new parameter type is exact.
- **No build or typecheck was run.** `source/node_modules` does not exist in this worktree and the instructions were not to run `pnpm install`. Verification is by careful reading only.
- Not run: unit tests, panel build, browser smoke test.


zbyju (commented):
**Verdict: No blockers**

Checked the three risk areas called out in the body, plus the surrounding code. All hold.

**Callers.** `grep` over the repo confirms exactly one import and one call site, both in `source/apps/panel/src/logic/panel/panelLoop.ts` (lines 27 and 103), and both are updated. The only panel spec file is `source/apps/panel/src/utils/panel/dataDisplay.spec.ts`; no test touches `getNewState`. The signature change has no other blast radius.

**Threshold fallbacks.** `source/apps/panel/src/logic/panel/state.ts:16-17` keeps the same `||` expressions, only the operand source moved. `0` still falls back (to 5 / 25) and so does `undefined`. `UnitsConfig` in `source/apps/panel/src/types/panel/config.types.ts:40-47` declares both as required `number`, and `getDefaultUnitsConfig()` in `source/apps/panel/src/defaults/config.default.ts:54-63` seeds them at 5 and 25, so the parameter is never `undefined` in practice either.

**Live config updates.** Not snapshotted. `this.unitsConfig` is the `storeToRefs(configStore).units` ref set once in the constructor (`panelLoop.ts:53,57`), but `updateState()` dereferences `.value` per tick at `panelLoop.ts:107`. `configStore.setConfig()` reassigns `this.units` wholesale, and the `storeToRefs` ref reads through to the store property, so a threshold change from the settings screen lands on the next tick. Same as before the change.

**Other things I looked at and cleared.**

- Removing `useConfigStore()` from `getNewState` cannot break Pinia context: the only caller already resolves the store in the `AppManager` constructor, inside a component setup.
- No lost reactivity. The old body read the store property inside a `setInterval` callback with no active effect, so nothing was ever being tracked there. Dropping the read changes no dependency graph.
- `this.unitsConfig.value` is a reactive proxy, not a plain clone. `getNewState` only reads two numbers off it and never mutates, so passing the proxy is harmless.

**PR body accuracy.** The claims I could cheaply check are correct: one call site, no test imports, `UnitsConfig` field types exact, fallbacks preserved, thresholds read live. The performance numbers are labelled as estimates and the "no build or typecheck" limitation is disclosed.

zbyju:

The three things checked are the three that mattered — caller completeness, identical `|| 5` / `|| 25` fallback semantics on `0` and `undefined`, and that `.value` is read per tick rather than snapshotted so live settings edits still land. Nothing to fix, no PR body corrections.

PR #68, opened 2026-09-09
perf: fetch panel config only once during boot

zbyju:
## What / Why

`initializeGlobal()` polls every 5s until the backend answers. Each tick also re-fetched the config and versions from configer and re-wrote both stores, even after the config had already loaded successfully.

In the common boot order (configer up, backend not yet), that meant two extra HTTP requests and a full invalidation of every config-dependent computed every 5 seconds. Config does not change during boot, so it is now fetched once.

## Performance impact

Estimated — not measured, no benchmark was run.

Per 5s boot tick, while the backend is still down:
- HTTP requests: 3 -> 1 (drops `GET /main` and `GET /version`, keeps the backend status call).
- Store writes: 2 -> 0 (`configStore.setConfig` and `versionsStore.setVersions` no longer run repeatedly).
- Reactive invalidation: every computed reading `configStore.units` / `configStore.app` / versions was re-evaluated once per tick; now zero after the first success.

Bundle size is unchanged apart from a few bytes (one local `let` plus a comment).

## Risk

Low.

The only behaviour change is skipping `initializeConfig()` after it has succeeded once. All four app-state transitions are preserved: `setConfigSuccess` fires exactly once on the first success, `setConfigError` still clears the interval and gives up permanently on failure, and `setBackendSuccess` / `setBackendError` are untouched and still run on every tick.

The boot screen cannot get stuck, because `appStateStore.done[0]` stays `true` once set and `setBackend*` calls `checkState()` on each tick, so a later backend success still flips the app to `Ok` and a later backend failure still shows `Trying`.

## Verification

- Read the full control flow in `source/apps/panel/src/logic/panel/panelLoop.ts` and `source/apps/panel/src/pinia/appStateStore.ts` and traced the boot orderings by hand: config ok + backend down, config ok + backend up, config fails.
- Did NOT run typecheck, build, or tests. `source/node_modules` does not exist in this worktree and installing was out of scope for this change. No build was run, so no build is claimed to pass.
- The change adds one boolean local and an `if` guard, so it introduces no new types or imports.


zbyju (commented):
**Verdict: No blockers**

Traced the boot control flow in `source/apps/panel/src/logic/panel/panelLoop.ts:179-207` against `source/apps/panel/src/pinia/appStateStore.ts` and `source/apps/panel/src/components/AppState.vue`. The guard is placed correctly and every path checked out.

What I checked:

- **Interval clearing.** Both clear sites survive: config failure clears at line 187, backend success clears at line 192. The only path that keeps the interval alive is config ok + backend down, which is exactly what main did. No new leak.
- **Config failure is still terminal.** `configLoaded` stays `false` on the error branch, but the interval is already cleared in the same statement, so there is no retry and no spin. `setConfigError` still sets `done[0] = false` and `state = Error`.
- **Success does not spin.** After the first success the tick body is one `initializeBackend()` call, unchanged.
- **Boot state machine still reaches Ok.** `done[0]` is never reset, and `setBackend*` calls `checkState()` on every tick, so config ok + backend down + later backend recovery still flips to `Ok`. Backend failure after config success still shows `Trying`. `configStore.initialised` (read by `AppState.vue:38`) is set by the one config write and stays true.
- **Versions are not skipped.** `versionsStore.setVersions` runs inside `initializeConfig` on the first success (line 154), which is the run that is kept.
- **await / `.then()` interleaving.** `configLoaded = true` is set inside the `.then()`, and the whole config block is `await`ed before `initializeBackend()` fires, so the guard is written and read in the same order it was before.

PR body claims verified: 3 -> 1 requests per tick is right (`getConfig` + `getVersions` + `getStatus`); 2 -> 0 store writes is right.

One non-defect note, no action needed: the callback is async, so if a config fetch ever exceeded 5s a second tick could start before `configLoaded` is set and fetch config twice. That is no worse than main, which refetched unconditionally.

zbyju:
Review addressed: **no code changes needed.**

Agreed on the non-defect note about the async interval callback: a config fetch slower than 5s can overlap ticks. Worth stating plainly that this is unchanged from `main`, where the overlap existed too and refetched unconditionally on every tick — this PR strictly reduces the number of overlapping fetches rather than introducing the pattern. Not fixing it here; it belongs with a wider boot-sequence cleanup, not a one-line perf guard.

PR #69, opened 2026-09-09
perf: drop unused recent-results queue from ConnectionTracker

zbyju:
## What / Why

`ConnectionTracker` kept a 999-entry `recentQueue` per unit and called `Array.prototype.shift()` on it once full — an O(n) re-index on every poll, for both units, every ~2s, forever on a kiosk that never restarts.

Nothing reads that queue. This removes `recentQueue`, `RECENT_SIZE` and the six queue-only readers (`getLatestResults`, `getAllLatestResults`, `countRecentSuccess`, `countRecentFails`, `getRecentQuality`, `getStats`). The counters the UI actually uses are untouched.

## Performance impact

Estimated — not measured on device.

- **Per tick:** removes 2 array pushes plus, once the queue fills (~33 min of uptime), 2 x `shift()` over 999 elements. That is ~2000 element moves per 2s tick that no longer happen, ~86M element moves per day, forever.
- **Memory:** 2 x 999-entry arrays of a numeric enum, held for the life of the process, freed. Small in bytes, but they were live reachable state on a device that never restarts.
- **Bundle:** ~55 source lines removed; a few hundred bytes minified. Not the point of the change.
- The remaining `addResult` work is four integer updates, O(1).

## Risk

**Low.**

The only behaviour that could break is the connection-quality display. The UI reads `requests`, `successes`, `fails`, `failStreak` (`logic/panel/tables.ts`, `logic/panel/state.ts`) and `getQuality()`, which is computed from the lifetime `requests`/`successes` counters, not from the queue. All four counter updates in `addResult` are byte-for-byte unchanged, and `getQuality()` / `calculateQuality()` are kept.

`ConnectionTracker` instances live in a Pinia store (`pinia/connectionStore.ts`) as class instances, so Pinia wraps them in a reactive proxy and every own property is tracked state. `recentQueue` was one of those tracked properties, but no template, computed or getter ever read it, so it had no reactive subscribers — dropping it cannot invalidate a render that was previously happening. The properties that do have subscribers (`requests`, `successes`, `fails`, `failStreak`) still exist, are still assigned in the constructor, and are still mutated in the same order. There is no persistence or serialization of this store (no `persist` plugin, no `JSON.stringify`, no `localStorage` anywhere in the panel), so no stored shape depends on the removed fields.

## Verification

**Dead-code check — done, and it is confirmed dead.** Grepped the whole repo (all apps, `.ts` / `.vue` / `.js` / `.json` / `.html`, script and template blocks, and outside `source/`) for `recentQueue`, `RECENT_SIZE`, `getStats`, `getLatestResults`, `getAllLatestResults`, `countRecentSuccess`, `countRecentFails`, `getRecentQuality`, and each `getStats` output key (`recentRequests`, `recentSuccesses`, `recentFails`, `recentQuality`). Every hit for all of them was inside `connections.ts` itself. Also grepped for serialization and persistence of the store — none exists. `getQuality()` and `calculateQuality()` are genuinely live (`tables.ts:67-68`) and were kept.

**Build / typecheck — NOT run.** `source/node_modules` is absent in this worktree and I was instructed not to run `pnpm install`. Verified by reading instead: the diff is deletions only, adds no new identifier, and no surviving line in the file or elsewhere in the repo references a removed member.


zbyju (commented):
**Verdict: No blockers**

Pure deletion, and the deleted code is genuinely dead. Checked independently rather than trusting the PR body.

**Dead-code check.** Grepped the whole repo (all file types, script and template blocks, inside and outside `source/`, node_modules excluded) for `recentQueue`, `RECENT_SIZE`, `getStats`, `getLatestResults`, `getAllLatestResults`, `countRecentSuccess`, `countRecentFails`, `getRecentQuality`, plus the `getStats` output keys `recentRequests` / `recentSuccesses` / `recentFails` / `recentQuality`. Every hit was inside `source/apps/panel/src/logic/panel/connections.ts` itself. No template-only usage.

**Surviving readers still work.**
- `source/apps/panel/src/logic/panel/tables.ts:67-68` calls `getQuality()`; it is computed from `this.requests` / `this.successes` via the retained `calculateQuality()`, never from the queue. Unaffected.
- `source/apps/panel/src/logic/panel/tables.ts:40-63` reads `requests` / `successes` / `fails` directly. Fields kept, still set in the constructor.
- `source/apps/panel/src/logic/panel/state.ts:181` sums `failStreak` for the "Chyba spojení!" alert. `addResult`'s four counter updates are byte-for-byte identical in the diff — same branches, same order, same reset of `failStreak` on success. Alert behaviour is unchanged.
- `source/apps/panel/src/logic/panel/panelLoop.ts:130-131` reads `.requests`. Unaffected.

**Store shape.** `source/apps/panel/src/pinia/connectionStore.ts:12-15` — the `connection` getter returns a new object holding the two tracker *instances*, not a spread of their fields, so no removed property was ever copied out. `source/apps/panel/src/types/panel/connection.types.ts` types the store by the class, so the removed members drop out of the interface with no dangling reference. No persistence: grep for `localStorage`, `sessionStorage`, `JSON.stringify`, `persist` across `apps/panel/src` returns nothing, so the PR body's serialization claim holds.

**Reactivity.** The removed fields had no reader, therefore no subscriber, so removing them cannot cancel a render that used to fire. The subscribed fields all remain.

**Body claims spot-checked.** `tables.ts:67-68` line reference is right. ~33 min to fill (999 ticks x 2s) and ~86M element moves/day (43,200 ticks x ~1,996 moves) both check out.

Note this is desk-checked only — the branch could not be checked out here (another worktree holds it) and no build was run, per the batch's known constraints. The diff is deletions only and adds no identifier, so the risk of that is low.

zbyju:

The independent grep of all 8 removed identifiers plus the 4 `getStats` output keys — including `<template>` blocks — is the check this PR needed, since a template-only usage was the one way a pure deletion could break the panel silently.

Noted that `gh pr checkout` collided with another worktree holding this branch, so this was a desk check against the diff rather than the branch tree. For a deletion-only diff with no new identifiers that is sufficient.

PR #70, opened 2026-09-09
perf: preload and reuse alarm sound instances

zbyju:
## What / Why

`playSound` built a new `Howl` on every playback, and Howler fetches and decodes the MP3 at construction time. Every alert therefore waited on a network round trip plus a decode before any sound came out.

The Howls are now built once, preloaded, and reused, so `playSound`/`stopSound` only call `play()`/`stop()` on already decoded audio. This is safety latency: `Aktivace` (babybox activated) and `Otevirani` (doors opening) are exactly the alerts that must not be late.

## Performance impact

- **Alert latency:** the fetch + decode moves from the alert path to app start. Estimated saving per alert: the local HTTP round trip plus MP3 decode of a 47-126 KB file. On a healthy LAN that is roughly tens of milliseconds; on a slow or briefly congested link it is unbounded. After the change the alert path is a `play()` call on a decoded buffer, so sub-millisecond. Not measured on real hardware.
- **Network:** total bytes fetched drop from `440 KB per full alert cycle, repeated forever` to `440 KB once per page load`. A panel that raises 50 alerts a day used to refetch and redecode a file each time.
- **Memory:** 5 decoded buffers held for the life of the page instead of one built and dropped per alert. Bounded and small (~440 KB compressed; decoded PCM is larger but still a few MB at most).
- **Bundle size:** unchanged, no new dependency.

## Design choices

**Preload the fixed set, and also cache lazily by name.** Both, not either.

The five files in `public/sounds/` are preloaded eagerly at module load, so the alerts that actually exist are hot before the first poll returns. But `updateSound` is driven by a `string` from `logic/panel/state.ts`, and nothing in the type system stops a new name being added there later. So `getHowl` falls back to building and caching any unknown name on first use. A name outside the fixed list therefore behaves exactly as it does today (it plays if the file exists), it does not throw and does not go silent. Preload-only would have made a future name silent; lazy-only would have left the first play of each alert slow, which is the whole problem.

**The cache is module level, not per player.** `useSounds()` runs on every `MainView` mount, and `MainView` is a plain route component (`router/index.ts`, path `/`) with no `<keep-alive>`, so every navigation back from `/settings` or `/data` remounts it. A per-instance cache would rebuild all 5 Howls on each of those, which is worse than today. Module-level means the set is built once per page load, no matter how many times maintenance staff navigate around.

**Failed loads are evicted.** With one-shot Howls, a transient fetch failure fixed itself on the next play. With cached instances it would not, so a `loaderror` handler drops the broken Howl and the next alert builds a fresh one. Without this, one bad fetch at boot would silence that alert until the next page reload, which is a real regression on a 24/7 kiosk.

Other behaviour kept as-is:
- The empty/null/undefined name guard in `loadHowl` is unchanged and still runs before any lookup.
- `stopSound()` on unmount still stops whatever the player last started.
- Preloaded Howls keep `autoplay: false`, so a never-played sound just sits decoded, it does not fire.
- `playSound` calls `stop()` on the target before `play()`, because a reused instance keeps the playhead from its previous run and could otherwise resume mid-file or stack a second concurrent playback.

## Risk

**Low.**

The two things that could break are shared state and startup cost. Shared state: because instances are reused, two `BabyboxSoundPlayer` objects would drive the same `Howl` — but only one `MainView` is mounted at a time and its unmount hook calls `stopSound()`, and `playSound` stops the target before playing, so overlapping playback is not reachable in practice. Startup cost: 5 preload fetches now happen at app load instead of on demand; they are ~440 KB from the same origin that serves the panel bundle, they are async, and nothing blocks on them (a play that arrives before a load finishes is queued by Howler and fires on load, same as the old behaviour).

## Verification

Honest account:

- **No build, no typecheck, no test run.** `source/node_modules` does not exist in this worktree and the task forbids `pnpm install`, so nothing was compiled or executed.
- **Verified by reading:** the diff is one file (`source/apps/panel/src/composables/useSounds.ts`, +39/-5); no `package.json` or lockfile touched. All call sites of `useSounds`/`playSound`/`stopSound`/`updateSound` were checked (`MainView.vue` only) and the sound names in `logic/panel/state.ts` (`BylOtevren`, `Otevirani`, `Aktivace`, `ZtrataSpojeni`) were confirmed against the files in `public/sounds/`. `MainView`'s mount behaviour was confirmed from `router/index.ts`. The `Howl.loop(boolean)` and `Howl.on("loaderror", ...)` signatures were checked against the declared `@types/howler` `^2.2.7`, not against installed types.
- Note for reviewers: the panel typecheck is already red on `main` with ~17 pre-existing errors, so a red `pnpm typecheck` on this branch is not by itself evidence of a regression from this change.
- **Not verified:** actual audio playback on the kiosk. Worth a manual smoke test — trigger an alert, navigate to `/settings` and back, trigger it again, and confirm the sound still plays and still stops.


zbyju (commented):
**Verdict: No blockers**

Defect-only review. Checked the silence risks against the actual Howler source (`howler@2.2.4`, matching the declared `^2.2.3`), not from memory.

**Play-during-load (the one that worried me most).** `playSound` now calls `stop()` then `play()` on the *same* instance. If an alert lands before preload finishes, `Howl.stop()` and `Howl.play()` both push onto `_queue` (`state !== 'loaded'`), giving `[stop, play]`. `_loadQueue` only advances when the matching event fires, so a `stop` that emitted nothing would strand the queued `play` and the alert would be silent. It does not happen here: `stop()` in 2.2.x guards on `if (sound)`, not `if (sound && !sound._paused)`, so it emits `stop` for the pool sound that `play()` created, `_loadQueue('stop')` shifts, and the queued `play` runs. `ZtrataSpojeni` firing during a slow preload — the nastiest coupling in this change, since the alert and the fetch share the same bad link — is safe for this reason. Double-`stop` (line 50 plus the new one, same name replayed) drains the same way.

**Replay correctness.** `stop()` resets `_seek` to `_start || 0`, so a reused instance restarts from the beginning, not mid-file. No truncation, no stacked playback.

**Loop flag.** `buildHowl` sets `loop: true`; `loadHowl` re-applies `howl.loop(loop)` on every play, and `loop(boolean)` writes both `_loop` and every pooled `sound._loop`. `playSound`'s only caller passes the default `true`. No path loses the loop flag.

**Unmount / remount.** `MainView` is the sole caller; `onBeforeUnmount` calls `stopSound()` on the shared instance, and a remount starts with `playing = null`. Only one `MainView` exists at a time, so no orphan keeps playing and no double-play.

**`loaderror` eviction.** Handler does not call `getHowl`, so no recursion. `loaderror` is load-time only under Web Audio, so it cannot evict a playing instance. A stale `this.playing` pointing at an evicted Howl is harmless — `stop()` on it is a no-op.

**Import-time preload.** `new Howl` does not throw, `autoplay: false`, and `import { Howl }` already ran Howler's global setup on `main`. Cannot take the panel down at boot.

**PR body claims verified.** No `<keep-alive>` anywhere in `src/`; `MainView` is a plain `/` route. The five preloaded names match `public/sounds/` exactly and total 443 KB. The four names in `logic/panel/state.ts` are as stated. The empty/null guard in `loadHowl` still runs before any lookup.

zbyju:

The play-during-load trace is the one that mattered and it is the exact failure this change could have introduced: `stop()` then `play()` queued on a still-preloading Howl would have stranded the `play` and silenced an alert. Verifying against the real `howler@2.2.4` source that `stop()` guards on `if (sound)` rather than `!sound._paused` — so the queue advances and the `play` fires — is what makes the `ZtrataSpojeni`-during-preload case safe.

No code changes, no PR body corrections.

PR #71, opened 2026-09-09
perf: import only the lodash functions the panel uses

zbyju:
## What / Why

The panel imported the whole lodash build (`import _ from "lodash"`) in two files but only ever called `_.isEqual` and `_.throttle`. Because the import pulls the monolithic entry point, Rollup cannot tree-shake it and the full library lands in the bundle.

This swaps the two default imports for deep per-function imports (`lodash/isEqual`, `lodash/throttle`) and updates the two call sites. `lodash` stays a dependency; nothing else changes.

## Performance impact

**Measured.** Both sides were built with `npx vite build` in a checkout with a real `node_modules`. Base is `main@bfae1ac`; head is this branch. Sizes are the panel's main JS asset in `source/apps/panel/dist/assets/`, raw bytes and gzip.

| build | raw | gzip |
|---|---:|---:|
| `main@bfae1ac` | 321,802 | 111,288 |
| this branch | 262,318 | 90,304 |
| **delta** | **−59,484 (−18.5%)** | **−20,984 (−18.9%)** |

`grep -c 'lodash.com/license' dist/assets/*.js` returns `1` on main and `0` on this branch, so the monolithic lodash build is gone rather than merely smaller.

For the record: an earlier version of this body carried an estimate of 55–60 KB minified / 18–20 KB gzip, made without `node_modules`. The measurement lands at 59.5 KB / 21.0 KB, so the estimate held, but the numbers above are the real ones and replace it.

No runtime effect. `isEqual` runs once per poll tick (~every 2 s) and `throttle` once per clock tick; both are the same functions, same implementations, just reached directly. This is load-time only, which still matters for a kiosk that reloads on restart.

## Risk

**Low.**

`lodash/isEqual` and `lodash/throttle` are the exact modules the monolithic build re-exports — same code, so behaviour is identical. The only realistic failure mode is a TypeScript interop complaint about the default import, and that risk is ruled out: `@types/lodash` uses `export =` for both the root entry and the per-function modules, so the deep imports use the same interop path the existing `import _ from "lodash"` already relied on. `tsconfig.app.json` extends `@vue/tsconfig/tsconfig.web.json`, which sets `esModuleInterop: true`.

Grep confirms no other lodash usage anywhere in `source/apps/panel/src` — the two lines in this diff are the only ones, and no residual `_.` call sites remain.

## Verification

What was and was not done:

- **Build comparison: done.** `npx vite build` on `main@bfae1ac` and on this branch, in a checkout with a real install. Numbers above. Also confirmed the lodash license banner is absent from the new bundle.
- **Typecheck: not run.** Worth knowing why this matters less than it looks, and more: `pnpm build` runs `vue-tsc --noEmit` against `tsconfig.json`, which has `"files": []` and only project references — and `vue-tsc --noEmit` without `--build` ignores references, so the build's type gate checks **zero files**. On top of that, `vue-tsc -p tsconfig.app.json` aborts with two `TS5102` config errors (`importsNotUsedAsValues` / `preserveValueImports`, both removed in this TS version) before it checks anything. Both problems are pre-existing on `main` and out of scope here, but they mean CI would not have caught an interop mistake in this diff. Type safety for this change rests on the reviewer's manual interop check (`export =` shape of `@types/lodash`, `esModuleInterop: true`), not on the type gate.
- **Runtime smoke test: not run.** Nobody loaded the panel in a browser against live units. The reviewer did confirm out-of-band that `lodash/isEqual.js` and `lodash/throttle.js` resolve in the installed package and each export a function directly.
- **Reviewed by hand:** 4 lines changed across 2 files, no other edits, no `package.json` or lockfile changes. `throttledBlink` is still created once per `useBigClockColon()` call, outside the `watch` handler, so throttle identity and behaviour are unchanged.


zbyju (commented):
**Verdict: No blockers**

Checked the four changed lines against a real `node_modules` (symlinked in from the main checkout, removed afterwards; no repo files touched).

What I verified:

- **No leftover `_` references.** `grep -rn "lodash|\b_\." source/apps/panel/src` on the PR head returns exactly the two new import lines and nothing else. Both call sites are updated. The PR body's grep claim holds.
- **`throttle` identity is preserved.** `throttledBlink` is still created once per `useBigClockColon()` call at `source/apps/panel/src/composables/useBigClockColon.ts:37`, outside the `watch` callback. The `watch` handler calls the same instance, so throttling still works. No behaviour change to the blinking colon.
- **`isEqual` call is a straight substitution** at `source/apps/panel/src/logic/panel/panelLoop.ts:108`. Same function, same arguments, same guard around `setState`.
- **TypeScript interop is fine.** `tsconfig.app.json` extends `@vue/tsconfig/tsconfig.web.json`, which sets `esModuleInterop: true`. `@types/lodash/isEqual.d.ts` and `throttle.d.ts` are both `export =`, so the default-import form is valid — the same path the old `import _ from "lodash"` used. Compiling just the two changed files with those settings produces no lodash error (the only error is a pre-existing `Cannot find namespace 'NodeJS'` at `panelLoop.ts:30`, unrelated to this diff and caused by my minimal `types: []` config).
- **Runtime resolution works.** `lodash/isEqual.js` and `lodash/throttle.js` exist in the installed package and both export a function directly (`throttle(...)` returns a cancellable function; `isEqual({a:1},{a:1})` is `true`).

Note: the repo's `vue-tsc -p tsconfig.app.json` aborts before checking any file with two `TS5102` config errors (`importsNotUsedAsValues` / `preserveValueImports` removed). That is pre-existing on `main` and not this PR's problem, but it means the type gate would not have caught an interop mistake here — worth knowing when relying on CI for the rest of this batch.

On the bundle estimate: the reasoning is sound in direction and rough magnitude (full lodash is ~71 KB min / ~24-25 KB gzip; the two subtrees are a small fraction of that). It is an estimate, clearly labelled as one, and the real number is being measured separately. Nothing to flag.

zbyju:
**Measured, replacing the estimate in the PR body.**

Built both `main@bfae1ac` and this branch with `npx vite build` in a checkout that has a real `node_modules` (the implementing agent's worktree had none, hence the estimate).

| | raw | gzip |
|---|---|---|
| main @ bfae1ac | 321,802 | 111,288 |
| this branch | 262,318 | 90,304 |
| **delta** | **−59,484** | **−20,984 (−18.9%)** |

`grep -c 'lodash.com/license' dist/assets/*.js` returns `0` on this branch and `1` on main, so the monolithic build is genuinely gone rather than merely smaller.

The PR estimated 55–60 KB minified / 18–20 KB gzip. Actual is 59.5 KB / 21.0 KB — the estimate was right.

zbyju:
Thanks — findings confirmed, all four match what is in the diff:

- No leftover `_` references; the two new import lines are the only lodash hits in `source/apps/panel/src`.
- `throttledBlink` is created once per `useBigClockColon()` call at `useBigClockColon.ts:37`, outside the `watch` handler, so throttle identity holds.
- `isEqual` at `panelLoop.ts:108` is a straight substitution.
- `esModuleInterop: true` via `@vue/tsconfig/tsconfig.web.json` plus `export =` in `@types/lodash` makes the default-import form valid.

No code changes made — none were asked for and none are needed.

**The estimate in the PR body is gone, replaced by measurement.** Both sides built with `npx vite build` against `main@bfae1ac`:

| | raw | gzip |
|---|---:|---:|
| main @ bfae1ac | 321,802 | 111,288 |
| this branch | 262,318 | 90,304 |
| **delta** | **−59,484 (−18.5%)** | **−20,984 (−18.9%)** |

`grep -c 'lodash.com/license' dist/assets/*.js` is `1` on main and `0` here, so the monolithic build is actually gone. The old estimate said 55–60 KB min / 18–20 KB gzip; actual is 59.5 KB / 21.0 KB.

On the `TS5102` note — it is worse than just that abort, and I put the full picture in the body's Verification section. `pnpm build` runs `vue-tsc --noEmit` against `tsconfig.json`, which is `"files": []` plus project references; without `--build`, `vue-tsc` ignores references, so the build's type gate checks zero files. The `TS5102` abort on `tsconfig.app.json` is the second, separate hole. Both pre-existing on `main`, both out of scope for this PR, but together they mean the type safety of this change rests on your manual interop check rather than on CI. Worth a separate issue for the batch.

Still not done: no typecheck run and no runtime smoke test in a browser. The body now says so plainly.

PR #72, opened 2026-09-09
perf: fetch engine and thermal settings concurrently

zbyju:
## What / Why

`fetchSettings()` awaited the engine unit, then awaited the thermal unit. They are two
physically separate devices on separate IPs with no shared state, so the serialisation
bought nothing. Both requests now start before either is awaited, so `unit=both` costs
one round trip instead of two.

## Performance impact

Estimated — not measured against real hardware.

- `unit=both`, happy path: latency drops from `engine + thermal` to `max(engine, thermal)`.
  On a healthy LAN both are a few ms, so the win is small in absolute terms.
- `unit=both`, one unit unreachable: worst case drops from 2 x timeout to 1 x timeout.
  With the default 5000 ms fetch timeout that is 10 s -> 5 s. Both units down: still 5 s
  instead of 10 s.
- `unit=engine` or `unit=thermal`: unchanged, one request as before.
- No extra requests are issued in any case: each unit is still fetched at most once.

## Error contract

Both requests are started up front, then each is wrapped in `Promise.allSettled` on its
own so its handler is attached in the same synchronous tick. Engine is awaited and
reported first. What the frontend sees is unchanged:

- Engine fails -> HTTP 500, "There was an error when fetching settings from engine unit."
- Engine ok, thermal fails -> HTTP 500, "...from thermal unit."
- Success shape is still `{ engine: <raw>, thermal: <raw> }`, with `null` for a unit that
  was not requested.

Latency on the error paths is also unchanged, and that took a second pass to get right.
The first version of this change awaited both results together
(`await Promise.allSettled([enginePromise, thermalPromise])`), which kept the status codes
and messages identical but *not* the timing: an engine that fails fast (connection
refused, ~1 ms) no longer answered until the thermal request had also settled. With a
thermal unit that hangs to the full 5000 ms timeout, the endpoint went from ~1 ms to
~5002 ms.

That is not invisible to the caller. The panel's own client timeout is also 5000 ms
(`backend.requestTimeout` in `configer/configs/base.json`, mirrored by
`DEFAULT_REQUEST_TIMEOUT` in `panel/src/api/base.ts`), and its timer starts before the
backend's, so the client would abort first every time — not just sometimes. In
`SettingsForm.vue` an aborted request is an `AxiosError` with no `.response`, which
matches neither the `status === 0` nor the `status === 500` branch, so the operator would
get **no log line at all** where the old code showed the connection error.

The fix is to await the engine result on its own before awaiting thermal. Both requests
still run concurrently and both are still handled the moment they are created, but a
failing engine returns straight away as it did before. Latency is now less than or equal
to the old sequential code in every case.

Verified across the failure matrix (old sequential vs. this version), responses
byte-identical throughout:

| unit | engine | thermal | old | this PR | response |
|---|---|---|---|---|---|
| both | ok | ok | 9 ms | 5 ms | 200 |
| both | fast fail | ok | 2 ms | 1 ms | 500 engine |
| both | ok | fast fail | 5 ms | 4 ms | 500 thermal |
| both | fast fail | hangs 5 s | 1 ms | 1 ms | 500 engine |
| both | hangs 5 s | fast fail | 5000 ms | 5003 ms | 500 engine |
| both | fast fail | fast fail | 1 ms | 1 ms | 500 engine |
| both | hangs 5 s | hangs 5 s | 5003 ms | 5000 ms | 500 engine |
| engine / thermal only | — | — | unchanged | unchanged | unchanged |

The one real behaviour change that stays invisible to the caller: when the engine request
fails, the thermal request has already been sent and its result is discarded. That is a
plain `GET` against a settings endpoint, so it has no side effect on the unit.

`allSettled` was picked over `Promise.all` because `all` rejects on the first failure and
would have made the "engine error wins over thermal error" precedence awkward to keep;
`allSettled` also guarantees neither promise is left unhandled.

Unchanged: the `isInstanceOfGetUnitSettingsRequest` check and its 400 path, and the single
shared `timestamp` cache-buster computed once and used for both URLs.

## Risk

Low. The function is a settings read used by the settings form, not by the 24/7 monitoring
loop (`panelLoop.ts` hits the data endpoints, not this one). Status codes, messages and the
response shape are byte-identical to before, so `SettingsForm.vue` needs no change. The
plausible failure mode would be a unit that cannot serve two overlapping HTTP requests —
but these are two different devices, so they never see more than one request each.

## Verification

- Did NOT run `pnpm install` or touch `package.json` / the lockfile.
- Type-checked `apps/backend` against the repo's real `node_modules`
  (`tsc --noEmit -p tsconfig.json`). `fetchFromUnits.ts` is clean. The only errors under
  `src/` are the two known pre-existing ones: `src/index.ts:82` (TS2349, `open` has no
  call signatures) and `src/types/request.types.ts:81` (TS2839). The remaining output is
  the known ambient noise from `bun-types` and `@types/jsdom` in `node_modules`.
- Ran the failure matrix above as a standalone Node harness covering all four `unit`
  inputs (`engine`, `thermal`, `both`, omitted) against the old sequential code, the
  await-both version and this one. Responses identical in all 12 cases; no
  `unhandledRejection` fired in any variant.
- No runtime test against real hardware or a stub. The existing jest tests under
  `src/**/__tests__/` do not cover `fetchFromUnits.ts`, so they would not have exercised
  this change either way.


zbyju (commented):
**Verdict: No blockers**

Checked the four input cases against `source/apps/backend/src/fetch/fetchFromUnits.ts:42-101`.

**Response equivalence — confirmed.**
- `unit=engine`: `thermalPromise` is `null`. `Promise.allSettled` settles a non-promise as `{status:"fulfilled", value:null}`, so `thermal` is `null`. Verified on Node: `Promise.allSettled([null, Promise.reject(e)])` -> `[{status:"fulfilled",value:null},{status:"rejected",...}]`.
- `unit=thermal` and unit omitted (`= "both"` default at line 51): same, mirrored.
- The 400 path and `isInstanceOfGetUnitSettingsRequest` are above the changed block and untouched.
- `timestamp` is computed once at line 54 and closed over by `settingsUrl`, so both URLs get the same cache-buster. Same as before.

**Error precedence — matches.** Engine is checked first, so engine wins when both fail. Old code returned the engine message in that case too (it never reached the thermal block).

**No unhandled rejection.** Both promises are created and passed to `allSettled` in the same synchronous tick, so a handler is attached before either can settle. `allSettled` never leaves a rejection floating. This was the main thing worth checking for a process that runs for months, and it is clean.

**Frontend.** `SettingsForm.vue:164-165` reads `data.engine` / `data.thermal` and calls `.split("|")` on both. It never sends a `unit` param (`getSettings()` in `source/apps/panel/src/api/units.ts`), so it always hits the `both` path. Shape unchanged.

**Typecheck.** Compiled the branch version against the real `node_modules`: `fetchFromUnits.ts` is clean. The only errors are the two known pre-existing ones (`index.ts:82` TS2349 on `open`, `request.types.ts:81` TS2839). `target: ES2020` in the backend tsconfig pulls in the `es2020.promise` lib, so `allSettled` is typed.

**One accuracy note, not a blocker.** The body says the discarded thermal result is "invisible to the caller". Latency is not. Old code returned the engine 500 the moment engine failed; new code always waits for thermal to settle too. If engine fails fast (connection refused) while thermal hangs the full 5 s, the backend now answers in ~5 s instead of ~1 ms, and the panel's own 5 s client timeout (`backendApi().timeout`) may fire first, so the operator sees a generic failure instead of "engine unit". Needs both units unhealthy in different ways, affects the settings form only, and is still bounded by the old worst case. Worth a line in the body rather than a code change.

zbyju:
Thanks — the latency finding is correct, and it is worse than the review makes it sound. I fixed it in code rather than just correcting the body. Pushed as `336d767`.

## Verifying the timeout race

Both numbers are 5000 ms, and neither is overridden anywhere:

- Backend: `getSettings()` in `panel/src/api/units.ts` sends no `timeout` query param, so
  `fetchSettings()` falls through to `defaultFetchTimeout()` in `backend/src/fetch/constants.ts`.
  That reads `DEFAULT_FETCH_TIMEOUT`, which is not set in `apps/backend/.env`
  (only `NODE_ENV`, `PORT`, `API_PREFIX`), so it is 5000.
- Client: `backendApi().timeout` is `backend.requestTimeout`, which is 5000 in
  `configer/configs/base.json` and 5000 again in `panel/src/defaults/config.default.ts`.
  `DEFAULT_REQUEST_TIMEOUT` in `api/base.ts` is also 5000, so every path lands on the same value.

So it is not a race the backend can win. The client's timer starts before the backend's,
and the backend's response is at 5000 ms + handler overhead, so the client aborts first
**deterministically** in the engine-fast-fail + thermal-hangs case — not "may fire first".

## The consequence is silence, not a generic error

The review says the operator "sees a generic client-side error instead of the specific
'engine unit' message". Two corrections:

1. The operator never sees "engine unit" in the first place. `SettingsForm.vue:182-199`
   discards the backend's `msg` and logs its own fixed Czech string per status code, so
   both 500s already render as "problém s připojením k jednotkám".
2. On a client timeout the axios error has no `.response`, so it matches neither the
   `err.response?.status === 0` branch nor the `=== 500` branch, and there is no `else`
   inside the `instanceof AxiosError` arm. **No log line is written at all.** The Load
   button just does nothing.

So the regression is not "less specific message", it is "error message" -> "no feedback".
On an unattended kiosk diagnosed over remote maintenance, that is worth more than the
few ms the parallel fetch saves.

## Why a code change, not a body note

There is a clean minimal fix: wrap each promise in its own `allSettled` and await the
engine result before awaiting thermal.

```ts
const settle = <T>(p: T) => Promise.allSettled([p]).then(([r]) => r);
const engineSettled = settle(enginePromise);
const thermalSettled = settle(thermalPromise);

const engineResult = await engineSettled;
if (engineResult.status === "rejected") { /* 500 engine */ }

const thermalResult = await thermalSettled;
if (thermalResult.status === "rejected") { /* 500 thermal */ }
```

Net +11 lines, most of them comment. It keeps everything you confirmed: both requests
still start in the same tick, both still get a handler in that same tick (so still no
unhandled rejection), engine still wins the error precedence, and the response shape and
`null` placeholders are untouched. It only stops the engine result waiting on thermal.

## Failure matrix

Ran all four `unit` inputs against three implementations — old sequential, the
await-both version you reviewed, and this fix. Responses byte-identical in all 12 cases;
no `unhandledRejection` in any variant.

| engine | thermal | old | await-both | this fix |
|---|---|---|---|---|
| fast fail | hangs 5 s | 1 ms | **5002 ms** | 1 ms |
| fast fail | ok | 2 ms | 3 ms | 1 ms |
| fast fail | fast fail | 1 ms | 2 ms | 1 ms |
| hangs 5 s | fast fail | 5000 ms | 5002 ms | 5003 ms |
| hangs 5 s | hangs 5 s | 5003 ms | 5002 ms | 5000 ms |
| ok | ok | 9 ms | 4 ms | 5 ms |

The one you flagged is the 5002 ms cell. Latency is now less than or equal to the old
sequential code in every row, so the change is a strict improvement rather than a
trade.

## What I deliberately did not do

- Did not touch `SettingsForm.vue`. The missing `else` in its catch is a real gap — a
  client timeout logs nothing — but it predates this PR and applies to every call it
  makes, not just settings. Out of scope here; worth its own PR.
- Did not change either timeout value. Making the client timeout longer than the
  backend's would be the systemic fix for the whole class of problem, but it is a config
  change affecting every endpoint and needs its own discussion.
- Did not touch `package.json` or the lockfile, and did not run `pnpm install`.

## Typecheck

`tsc --noEmit -p tsconfig.json` on `apps/backend` against the real `node_modules`:
`fetchFromUnits.ts` clean. Same two pre-existing `src/` errors you found
(`index.ts:82` TS2349, `request.types.ts:81` TS2839) and nothing else. `Awaited<T>` and
the tuple-preserving `allSettled` overload the helper relies on are both present.

PR body updated — the "Error contract" section now documents the timing explicitly
instead of claiming the change is invisible to the caller.

PR #73, opened 2026-09-09
perf: lazy-load the settings and data routes

zbyju:
## What / Why

The router statically imported all three views, so Vite emitted the panel as one JS chunk with no code splitting. Settings and Data are now dynamic imports; MainView stays eager because it is the boot route.

Settings is a technician-only screen behind a password and Data is an 11-line stub, yet both were parsed and evaluated on every boot. The panel reloads itself periodically (`refreshRequestLimit` in `source/apps/panel/src/logic/panel/panelLoop.ts`), so this cost is paid over and over on a 24/7 kiosk, not once.

## Performance impact

**Measured, not estimated.** Both sides built with `npx vite build` in a checkout with a real `node_modules`, base `main@bfae1ac`. The earlier estimate in this body has been replaced by the numbers below.

Boot transfer — what the kiosk downloads before the panel renders:

| | JS gzip | CSS gzip | total gzip |
|---|---|---|---|
| `main@bfae1ac` | 111,288 | 3,167 | 114,455 |
| this branch | 107,394 | 2,469 | 109,863 |
| **delta** | −3,894 | −698 | **−4,592 (−4.0%)** |

Chunks emitted on this branch: `index` 107,394 gz, `SettingsView` 4,617 gz plus 1,243 gz of CSS, `DataView` 186 gz.

Two things the original estimate did not separate out:

- **CSS splitting is 15% of the win.** 698 of the 4,592 gzipped bytes are style, not script. Each lazy view's `<style lang="stylus">` block moves into its own CSS chunk that Vite only fetches with the route, so boot CSS drops from 3,167 to 2,469 gz. The estimate folded this into the JS figure.
- **Total bytes across all chunks go up slightly.** Boot 109,863 gz + the Settings chunks 5,860 gz = 115,723 gz, against 114,455 gz for the single bundle on `main`. Splitting costs about 1.3 KB in gzip efficiency. That is paid only by a technician who actually opens Settings; the kiosk, which never leaves MainView, only sees the 4.0% saving.

The PR originally estimated 4–6%. Actual is 4.0% — the bottom of the range.

**What does NOT move, to be clear about it:**

- **No npm dependency leaves the main chunk.** `moment`, `lodash`, `axios`, `howler`, `vue`, `pinia` and `vue-router` are all reachable from MainView's own tree — `moment` via `utils/time.ts` and `utils/panel/conversions.ts`, `lodash` via `logic/panel/panelLoop.ts` and `composables/useBigClockColon.ts`, `axios` via `api/units.ts`, `howler` via `composables/useSounds.ts`. The settings tree adds no external dependency the main view does not already pull in. The vendor bulk, which dominates the bundle, is untouched. This is why the saving is 4% and not 40%.
- What does move is app code: 17 modules, ~33.1 KB of source — `components/settings/**` (8 files), `utils/settings/*`, `defaults/settingsTable.defaults.ts`, the two views, and two type-only files that compile to nothing.
- `components/TheNav.vue` and `components/panel/HTMLElements/BaseInput.vue` stay in the main chunk — MainView imports `TheNav` directly, and `BaseInput` is shared with the panel tables.
- DataView is 186 gz on its own. It is included for consistency, not for the bytes.

Side benefit: `SettingsView.vue` imports `@/router`, which the router imported back. That static cycle is now broken at boot.

## Risk

**Low.**

Opening Settings now needs one extra HTTP round trip to fetch its chunk. If that fetch fails the route will not render, where previously the code was already in memory.

I do not think this bites in practice. The chunk is served by the same Express process that serves `index.html` (`express.static(__dirname + "/public/")` in `source/apps/backend/src/index.ts`), over the local network, from a server the panel is already polling every ~2s — if it is unreachable, the panel is already dead and the technician has a larger problem. MainView, the safety-critical view, is unchanged and still eager. `express.static` serves `/assets/*.js` directly and the SPA fallback is only `app.get("/")`, not a catch-all, so a stale chunk request returns a clean 404 rather than an HTML body with a JS MIME type.

Vite 2.9 with the default `modules` build target emits native dynamic imports, and Vue Router 4 has supported async route components since its first release, so no config change is needed.

**Stale-chunk window.** The deploy in `source/apps/startup/src/logic/start/ubuntu.js:97-141` renames `dist` to `dist2` before copying, so old hashed chunks are deleted. A page loaded before that deploy would request a chunk that no longer exists. The same flow then restarts the backend, which calls `open(...)` and loads a fresh `index.html`, so the stale page would have to survive a deploy without that reopen — and F5 recovers it.

**Known gap, deliberately not fixed here.** There is no `router.onError`, so a failed chunk fetch makes the Settings link do nothing at all, with no message. See Follow-up.

## Verification

Honest account of what was and was not run.

**Run:**

- `npx vite build` on `main@bfae1ac` and on this branch, in a checkout with a real `node_modules`, gzipped sizes compared. That is where the Performance impact table comes from. The build succeeds on both sides and the expected three chunks are emitted.

**Not run — and this is the gap that matters:**

- **No runtime smoke test.** Nobody has opened `/settings` on a built bundle and watched the chunk load. A broken dynamic import is exactly the failure a build does not catch: Vite will happily emit a chunk for a path that fails at runtime, and the route only executes when a human clicks the link. This is the single check worth doing before merge.
- **The panel's type gate is not a safety net here.** `build` runs `vue-tsc --noEmit`, but `source/apps/panel/tsconfig.json` is `{"files": [], "references": [...]}`, so that invocation typechecks zero files. `pnpm typecheck` (which points at `tsconfig.vitest.json`) is already red on `main` with ~17 pre-existing errors. Neither would flag a bad route import.
- No lint, no unit tests.

**Also checked, by reading:**

- `source/apps/panel/src/router/index.ts` and both views end to end; `../views/SettingsView.vue` and `../views/DataView.vue` resolve from `src/router/`.
- No other module imports either view, so nothing pulls them back into the entry chunk.
- `vite.config.ts` sets no `base`, no `rollupOptions`, and no `build.target` override that would interfere with code splitting.
- The diff is one file, 2 insertions / 4 deletions, no reformatting, no `package.json` or `pnpm-lock.yaml` change.

**Before merge:** run `pnpm -F babybox-panel preview` against the built `dist`, click Settings, confirm the chunk is fetched and the view renders, then navigate back.

## Follow-up (not in this PR)

`router.onError` is a real gap, but it does not belong in a 2-line routing change:

- The obvious one-liner, `router.onError(() => window.location.reload())`, is the wrong fix on a kiosk. It reloads the whole monitoring display — dropping panel state and restarting the poll loop — in response to a technician clicking a link. If the chunk is genuinely missing, every click reloads the display again and the technician still never reaches Settings.
- A correct handler needs a guard against repeat reloads and a visible message, which is not one line and needs its own test.
- The exposure is already bounded: the deploy restarts the backend and reopens a fresh `index.html`, and F5 recovers any page that slips through.

Worth a separate issue covering `router.onError` for the whole app, not a rider on this change.


zbyju:
**Measured, replacing the estimate in the PR body.**

Built both `main@bfae1ac` and this branch with `npx vite build` in a checkout that has a real `node_modules` (the implementing agent's worktree had none).

Boot transfer — what the kiosk downloads before the panel renders:

| | JS gzip | CSS gzip | total gzip |
|---|---|---|---|
| main @ bfae1ac | 111,288 | 3,167 | 114,455 |
| this branch | 107,394 | 2,469 | 109,863 |
| **delta** | −3,894 | −698 | **−4,592 (−4.0%)** |

Chunks emitted on this branch: `index` 107,394 gz, `SettingsView` 4,617 gz (+1,243 gz CSS), `DataView` 186 gz.

The PR estimated 4–6% of gzipped boot transfer. Actual is 4.0% — at the bottom of the range, and the PR was right to deflate the headline number and state that no npm dependency leaves the main chunk.

zbyju (commented):
**Verdict: No blockers**

Defect review of the 2/4-line router change. Nothing here breaks behaviour.

Checked and correct:

- Paths resolve. `router/index.ts` sits in `src/router/`, so `../views/SettingsView.vue` and `../views/DataView.vue` point at the two real files. No other module imports either view (grep over `src/`), so nothing pulls them back into the entry chunk.
- Types are fine. `RouteRecordRaw.component` accepts a lazy component, and `@vue/tsconfig/tsconfig.web.json` sets `module: esnext`, so the dynamic import needs no config change. Vite 2.9's default `build.target: "modules"` emits native dynamic imports.
- Import-cycle claim is true. `SettingsView.vue` does `import router from "@/router"` while the router imported the view back. That static cycle is gone.
- Dependency claim is true. `moment`, `axios`, `lodash`, `howler`, `vue`, `pinia`, `vue-router` are all reachable from MainView's tree (`logic/panel/panelLoop.ts`, `api/units.ts`, `utils/time.ts`, `composables/useSounds.ts`), so the settings tree removes no npm package from the entry chunk. The ~4-6% gzipped estimate is app-code only, which the body states.
- Deep-linking is unchanged. `GET /settings` 404s on `main` and on this branch alike, because the backend has only `app.get("/")` (`source/apps/backend/src/index.ts:83`). Navigation is client-side `router-link` from `components/TheNav.vue`, so the kiosk path is unaffected.

Non-blocking, for the record: there is no `router.onError` handler, so a failed chunk fetch makes the Settings link do nothing at all, with no message. The deploy flow limits the exposure — `source/apps/startup/src/logic/start/ubuntu.js:97-141` renames `dist` to `dist2` before copying, so old hashed chunks are deleted, but the same flow then starts the backend, which calls `open("http://localhost:" + port)` and loads a fresh `index.html`. A stale page would have to survive a deploy without that reopen, and F5 recovers it. Not worth blocking on; a one-line `router.onError(() => window.location.reload())` would close it if you want belt and braces.

zbyju:
Review points all land. Body rewritten: the estimate is gone, the measured build numbers are in, and Verification now says what was and was not run.

**On the measurement.** 4.0%, bottom of the estimated 4–6% range. Two things the estimate got wrong that are now called out separately:

- 698 of the 4,592 gzipped bytes are CSS, not JS — 15% of the win came from style splitting, which the estimate folded into the JS figure.
- Total bytes across all chunks go *up* by ~1.3 KB gz (115,723 split vs 114,455 single). Splitting costs gzip efficiency. Only a technician who opens Settings pays it; the kiosk does not.

**On `router.onError` — not doing it here, and not just for scope reasons.** Your one-liner is the wrong fix on this device. `window.location.reload()` reloads the whole monitoring display in response to a technician clicking a link: panel state gone, poll loop restarted. And if the chunk is genuinely missing it does not help — every click reloads again and Settings still never opens. A correct handler needs a repeat guard and a visible message, which is not one line and wants its own test. Recorded as a follow-up in the body, scoped to the whole app rather than as a rider on a 2-line routing change. Push back if you think the reload risk is overstated.

**Correcting one thing in the body that also affects your read of the type check.** The body previously implied the build's `vue-tsc --noEmit` would catch an import problem. It would not: `source/apps/panel/tsconfig.json` is `{"files": [], "references": [...]}`, so that invocation typechecks zero files. `pnpm typecheck` uses `tsconfig.vitest.json` and is already red on `main` with ~17 pre-existing errors. Neither gate would flag a bad route import.

**The remaining gap is the one worth acting on.** Nobody has opened `/settings` on a built bundle. A `vite build` emits a chunk for a dynamic import that fails at runtime just as happily as for one that works, and with no type gate behind it, a `preview` click-through is the only check that would catch a broken import. Flagged in the body as the thing to do before merge. Not claiming it as done.

PR #74, opened 2026-09-09
perf: refresh camera image only after the previous frame finishes

zbyju:
## What / Why

`useCamera` drove the snapshot `` from a free-running `setInterval(..., config.updateDelay)` (1000 ms by default), rewriting `src` with a fresh timestamp whether or not the previous snapshot had finished.

Dahua/Hikvision snapshot endpoints regularly take longer than 1 s under load. Writing `src` aborts the request already in flight, so on a slow camera the old code cancelled each snapshot before it could finish and **no frame ever arrived**. The refresh is now self-clocking: the next url is scheduled from the image's load/error completion, with `config.updateDelay` kept as the **minimum** gap measured refresh-to-refresh (a fast camera still updates at 1 Hz, exactly as before).

The gap is measured with `performance.now()`, not `Date.now()`, so an NTP step on the kiosk cannot move the next refresh.

## Performance impact

Estimated — not measured on real hardware.

- **Healthy camera (snapshot < 1 s):** no change. 1 request/s, 1 JPEG decode/s.
- **Slow camera (e.g. 2.5 s snapshot):** this is the real win, and it is a correctness win, not just a throughput one. Assigning `img.src` aborts the previous request, so the old code also had only **one** in-flight request at a time — but it tore that request down after 1 s and started over, so a 2.5 s camera delivered **zero completed frames**: the panel showed a permanently stale image while opening ~1 connection/s forever. After this change the request is allowed to finish, so the feed delivers ~0.4 frames/s where it previously delivered none, and opens ~60% fewer connections (~0.4/s instead of ~1/s), with no more cancelled TCP/TLS setup or wasted partial JPEG decode.
- **Dead/refused camera:** unchanged, ~1 retry/s.
- **Stalled connection (no load, no error):** before, 1 doomed request/s forever. After, 1 request per 15 s.
- Bundle size: negligible (a few hundred bytes of extra logic, no new dependency).

## Risk — Low/Medium

Medium-low. This is the panel's live camera feed on an unattended kiosk, so a scheduling bug means no camera until someone drives to the hospital. Specific hazards and why I think they are covered:

- **Feed stops after one error.** Handled: `imageFinished()` is called from *both* `onerror` and `onload` in `CameraView.vue`, so a dead camera retries forever at `updateDelay`, same as today.
- **Feed deadlocks when neither event fires** (stalled TCP, or an identical url that the browser does not reload). Handled by a 15 s watchdog per frame: if the frame never reports back, it is dropped and a new one starts. 15 s was chosen because a monitoring frame that old is already useless, but it is long enough not to cut off a genuinely slow camera and start the pile-up again. Replacing `src` cancels the abandoned request, which is the same cancel path the old code used — just at 15 s instead of 1 s.
- **A wall-clock step delays the next refresh.** Handled: the refresh-to-refresh gap is measured with `performance.now()`, which is monotonic and unaffected by NTP. With `Date.now()` a backward step of *n* would have pushed the next refresh out by *n* — and since `imageFinished()` clears the stall watchdog first, nothing would have recovered the feed. `Date.now()` is still used for the url cache-buster only, where a repeated value costs at most one watchdog cycle.
- **Regressing ada6ee9 ("clear camera interval on unmount").** Handled: `onUnmounted` sets a `stopped` flag *and* clears both the pending refresh timer and the watchdog timer. `stopped` also blocks any late `imageFinished()` from a DOM event that lands after unmount from scheduling a new timer.
- **Spurious first event.** `url` starts as `""`, and `` can fire load or error before the first refresh. `imageFinished()` no-ops unless a frame is actually in flight (`waitingForImage`), so this cannot start a second timer chain.
- **Signature change.** `useCamera` now returns `{ url, imageFinished }` instead of a bare `Ref<string>`. It has exactly one caller (`CameraView.vue`), updated here. `VivotekCameraView.vue` does **not** use the composable at all — it opens an iframe — so it is untouched. The optional `onUpdate` callback keeps its contract (called after every url change) and still has no callers.
- Vivotek path: `CameraView` calls `useCamera` even when it renders the Vivotek branch, where no `` exists. Before, that spun a useless 1 Hz timer; now it settles to one timer per 15 s. Strictly less work, no behaviour change (no requests are made either way).

## Verification

- `source/node_modules` does **not** exist in this worktree and I was told not to run `pnpm install`, so **I did not run a build, typecheck, lint, or test.** No build passed — I am not claiming one did.
- What I did run: `node --experimental-strip-types --check` on `useCamera.ts` (exit 0). That is a **parse check only**, not type checking.
- What I did by reading: traced every `useCamera` caller (`grep` across the repo — one caller), confirmed `VivotekCameraView.vue` does not use it, confirmed `Ref` is still imported and used in `useCamera.ts` and removed the now-unused `type Ref` import from `CameraView.vue`, and checked line width against `.prettierrc.json` (`printWidth: 80`).
- The clock-step arithmetic was reproduced in a standalone Node script, before and after the fix — not in the app.
- Not verified: real-camera behaviour, and the exact browser event ordering for ``. The `waitingForImage` guard is written so either ordering is safe.


zbyju (commented):
**Verdict: 1 blocker**

### 1. A backward wall-clock step freezes the feed for the size of the jump

`source/apps/panel/src/composables/useCamera.ts:88-89`

```ts
const sinceRefresh = Date.now() - lastRefreshAt;
scheduleRefresh(Math.max(config.updateDelay - sinceRefresh, 0));
```

The delay is clamped at the bottom but not at the top. `sinceRefresh` comes from two `Date.now()` reads, so an NTP step backwards between `refresh()` and `imageFinished()` makes it negative, and the delay becomes `updateDelay + jump`.

Failure scenario: kiosk RTC runs fast, chrony/ntpd steps the clock back one hour. A frame loads across the step. `imageFinished()` computes `1000 - (-3600000)` and arms `nextTimer` for ~60 minutes. It has already called `clearTimers()`, so the 15 s stall watchdog is gone — the one mechanism meant to recover an unattended feed does not cover this path. No camera for an hour, nobody on site to reload.

The old `setInterval` was immune: it never read the clock.

Fix is one line — also clamp the top: `Math.min(Math.max(config.updateDelay - sinceRefresh, 0), config.updateDelay)`.

### Checked and clean

- **No permanent stop otherwise.** Every path leaves exactly one live timer. `refresh()` always arms `stallTimer`; `imageFinished()` always reaches `scheduleRefresh` (the `stopped` guard already returned at line 85). Error, stall, and late-event paths all keep the chain alive.
- **No double-scheduling.** `scheduleRefresh` only runs from setup or straight after `clearTimers()`; `refresh` only runs from a timer that just fired or after `clearTimers()`. No orphaned timer, so no parallel chain. A stale event landing after a watchdog restart costs one early refresh, then self-corrects.
- **Handler ordering.** `onMounted` runs in the same task/microtask flush as setup; the first refresh is a `setTimeout` at 1000 ms, so the handlers always exist first. `v-show` (not `v-if`) keeps the `` in the DOM in all three states, so the handlers survive an error.
- **ada6ee9 not regressed.** `stopped` plus both `clearTimeout` calls; a late DOM event after unmount no-ops.
- **`VivotekCameraView.vue`** confirmed — no `useCamera`, uses `window.open` into an iframe. `onUpdate` still fires after every url change and still has no callers. One caller only.

### PR body nit (not blocking)

"roughly 2.5 in-flight TCP connections at all times" and "delivered frame rate is unchanged" are both wrong. Writing `img.src` aborts the previous request, so the old code also had one in-flight request — but a 2.5 s camera delivered **zero** frames, since each was cancelled after 1 s. The PR understates its own benefit.

zbyju:
Blocker confirmed and fixed in `0c15ab9`. I took a different fix than the one suggested — reasoning below.

### Verified the blocker independently

The failure is real. `lastRefreshAt` was written from `Date.now()` in `refresh()` and read back from `Date.now()` in `imageFinished()`, so a backward step lands between the two reads. I reproduced the arithmetic standalone (1000 ms `updateDelay`, 400 ms frame, clock stepped back one hour):

```
sinceRefresh = -3599600   ->   delay = 3600600 ms  (60.01 minutes)
```

And the recovery argument holds: `imageFinished()` calls `clearTimers()` before scheduling, so `stallTimer` is already gone and nothing re-arms it. The feed is dead for the size of the step with no watchdog behind it. It is a genuine regression against the old `setInterval`, which never read a clock.

### What I changed — `performance.now()`, not a top clamp

I did not use `Math.min(..., config.updateDelay)`. I switched the measurement to a monotonic clock:

```ts
lastRefreshAt = performance.now();
...
const sinceRefresh = performance.now() - lastRefreshAt;
```

Why this over the clamp:

- **It removes the bug rather than bounding it.** The clamp caps the damage at `updateDelay`, but `sinceRefresh` is still wrong after a step, so the gap between two refreshes is still wrong — just not catastrophically. `performance.now()` is monotonic by spec and unaffected by NTP, so the subtraction is correct in every case and the quantity means what the code says it means.
- **Same line count, no new state.** This is not the bigger change; it is the same size.
- **Forward jumps: considered, no handling needed.** A forward step makes `sinceRefresh` large and positive, so `updateDelay - sinceRefresh` goes negative and the existing `Math.max(..., 0)` floors it at 0 — one early refresh, then self-correcting. That was never a stall. With `performance.now()` the case disappears entirely, since no wall-clock step reaches the arithmetic at all.
- **Cannot reintroduce a stall.** With a monotonic source `sinceRefresh >= 0` always, so `updateDelay - sinceRefresh <= updateDelay`. The top clamp becomes unreachable code. I deliberately left it out rather than keep a dead guard that implies the primitive is not trusted.

`Date.now()` is still used for the url cache-buster, on purpose. A backward step there can repeat a timestamp, which means an unchanged `url` and no load/error event — but that path *is* covered by the 15 s watchdog (`refresh()` arms it before returning), so it self-heals in one cycle. Same behaviour as the old code. Out of scope to change here.

### Body nit — accepted, and it was understated in both directions

You are right and the correction is bigger than a wording fix. Since `img.src` aborts the in-flight request, the old code had one connection, not 2.5 — but on a 2.5 s camera it also cancelled every snapshot at 1 s, so it delivered **zero** completed frames while opening ~1 connection/s forever. "Delivered frame rate is unchanged" was the wrong claim: the frame rate went from 0 to ~0.4/s. The body now says this. That reframes the change from a throughput optimisation into a correctness fix for slow cameras, which is what it actually is.

### Unchanged

The six "checked and clean" points match what I read — single live timer on every path, no double-scheduling, handler ordering, `ada6ee9` not regressed, `VivotekCameraView.vue` not a caller. Nothing else touched.

Still no build or typecheck: `source/node_modules` is absent and I was told not to install. Parse check on `useCamera.ts` passes; the clock arithmetic was tested in Node, not in the browser.

PR #75, opened 2026-09-09
perf: add cache headers to the served panel assets

zbyju:
## What / Why

`express.static` served the built panel with `serve-static`'s default `maxage: 0`, which emits `Cache-Control: public, max-age=0` on every file. The header exists, but it means the browser revalidates every asset on each load — one conditional request per file. The panel reloads itself on purpose once `refreshRequestLimit` (default 50000 requests, roughly every 28h at a 2s poll) is hit, so full page loads repeat for the machine's whole life. This sets an explicit cache policy per path.

## Cache policy per path

| Path | `Cache-Control` | Why |
|---|---|---|
| `index.html` (static middleware **and** the `app.get("/")` handler) | `no-cache` | Always revalidated. A cached index would pin the kiosk to old code after a deploy until someone clears the browser cache on site. `no-cache` still allows a 304, so the cost is one small conditional request, not a re-download. |
| `/assets/**` (JS, CSS) | `public, max-age=31536000, immutable` | Vite content-hashes these names (`index.68d1c453.js`, `index.2d5c45e0.css`), so changed content always arrives under a new URL. Safe to never revalidate. |
| Everything else — `sounds/*.mp3`, `fonts/`, `favicon.ico`, `config/styles.json` | `public, max-age=86400` | Copied from `panel/public/` verbatim, so the names are stable and a long cache would mean serving an old copy. One day is short enough that a corrected alert sound reaches the kiosk on its own, and long enough to skip the re-download on a same-day reload. No `immutable`, so a revalidation after a day is still possible. |

Both `index.html` entry points are covered: `express.static` serves `/` first via its directory index and gets the header from `setHeaders`; the explicit `app.get("/")` fallback passes the same header to `sendFile`.

`config/styles.json` is listed above only because it sits in `public/` and is therefore copied into `dist/` and served. Nothing fetches it at runtime — `panel/src/assets/styles/variables.styl:3` reads it at **build** time via stylus `json(...)`, and a built bundle contains zero references to it. Its cache policy is inert.

## Performance impact

Estimated — not measured on a real kiosk. Asset sizes below are measured from `source/apps/panel/public/`.

Today, with `max-age=0`, a repeat load revalidates every file. Assets stay in the browser cache and normally come back `304 Not Modified`, so the cost is a round trip per file, not the bytes. Bytes are only re-transferred on a cold cache.

Per page load:

- `index.html` (~0.4KB), the hashed JS/CSS (~322KB), `favicon.ico` (894B), `fonts/opensans.css` (1.4KB), and the Open Sans TTFs. The 10 TTFs are **~1.0MB** in total (1,010,684 bytes), but they are `@font-face` declarations, so a browser only fetches the faces it actually renders — ~1.0MB is the ceiling, not the typical load.
- That is up to ~15 requests per load, fewer in practice.
- The alert MP3s are **not** part of a page load. `panel/src/composables/useSounds.ts` builds a `Howl` only inside `playSound`, so the 5 files (~0.44MB total) are fetched when an alert first fires, and again on each later play because a fresh `Howl` is constructed every time. The 24h policy still helps there, just not at load time.

After this change: hashed assets are served straight from the browser cache with zero requests; fonts, favicon and sounds are also cache hits within 24h. Only `index.html` hits the server, as a single ~0.4KB conditional request.

The panel normally runs against `localhost`, so the win is mostly request overhead, disk reads and parse setup, not network bytes. It matters more for a remote maintenance session over the hospital network. All served assets together are **~1.8MB** (~1.46MB in `public/` plus ~322KB of hashed JS/CSS).

## Risk

**Low.** Only the production static-serving branch changes; dev is untouched. The one real failure mode — a deploy that never reaches the kiosk — is exactly what the `index.html` `no-cache` rule prevents: the HTML is revalidated on every load, and it is the only file that names the hashed asset URLs, so new JS/CSS is picked up as soon as the new HTML is. The residual exposure is a stale sound, font or favicon for up to 24h after a deploy that changes one of those files without renaming it.

## Verification

- Ran the backend typecheck: `tsc -p tsconfig.json --noEmit --typeRoots ./node_modules/@types`, on this branch and on the base commit. Identical output both times — the change introduces no new errors. Two errors are pre-existing on `main` and unrelated: `request.types.ts:81` (TS2839 object comparison) and the `open()` call in `index.ts` (TS2349), the latter only because the `node_modules` I borrowed to compile has `open@10` rather than the `open@8` this branch's lockfile pins.
- Runtime headers confirmed by rebuilding the exact `setPanelCacheHeaders` + `express.static` + `sendFile` wiring against this repo's own `express@4.18.1` / `serve-static@1.15.0` and requesting each path. Base wiring returns `public, max-age=0` on `/`, `/index.html`, `/assets/*` and `/sounds/*`; the new wiring returns `no-cache`, `public, max-age=31536000, immutable` and `public, max-age=86400` on the paths in the table above. `If-None-Match` on `index.html` (both `/` and `/index.html`) returns `304`, so `no-cache` cannot pin the kiosk to stale code.
- Asset sizes measured directly from the checked-out `panel/public/` tree. `config/styles.json` confirmed absent from a built `dist/assets/*.js`.
- No `pnpm install` was run and no `package.json` / `pnpm-lock.yaml` was touched, per instructions.
- Not run: no panel build on this branch, so the exact hashed filenames in the table come from a reported build rather than one produced here. `assetsDir: "assets"` is the Vite default and a local build confirms the `dist/assets/` layout.


zbyju (commented):
**Verdict: No blockers**

I reproduced the exact `setPanelCacheHeaders` + `express.static` + `sendFile` wiring against `express@4.18.1` from this repo's store and curled it. The header table in the PR body is correct in practice:

| request | actual header |
|---|---|
| `/` | `Cache-Control: no-cache` |
| `/index.html` | `Cache-Control: no-cache` |
| `/assets/index.abc123.js` | `public, max-age=31536000, immutable` |
| `/sounds/Start.mp3` | `public, max-age=86400` |

Specific things checked, all clean:

- **The matcher fires on the right argument.** `serve-static` resolves the root, so `send` emits an absolute filesystem path. `path.basename(...) === "index.html"` and `startsWith(HASHED_ASSETS_DIR + path.sep)` both work on it. `HASHED_ASSETS_DIR` derives from the same `PUBLIC_DIR`, so no prefix mismatch. Confirmed by the real responses above, not by reading alone.
- **The header is not clobbered.** `send@0.18.0` `SendStream.prototype.setHeader` emits `headers` *first*, then sets its own `Cache-Control` only `if (!res.getHeader('Cache-Control'))`. The explicit header wins over the middleware default.
- **`app.get("/")` never runs.** `express.static`'s directory index answers `GET /` first, and `setHeaders` gives it `no-cache`. The `sendFile` fallback is dead code here, but it carries the same header if it ever fires (express 4 applies `options.headers` on the same `headers` event).
- **`no-cache` still revalidates.** ETag and `Last-Modified` are still set; `If-None-Match` returns `304 Not Modified`. No risk of pinning the kiosk to stale code.

Three factual corrections to the PR body (none change the verdict):

1. "*served the built panel with no `Cache-Control` at all*" is wrong. `serve-static` defaults `maxage` to 0, so main already sends `Cache-Control: public, max-age=0` on every file. I verified this against the base wiring. The described *behaviour* (revalidate every load) is right; the header does exist.
2. "*~2MB of Open Sans TTFs*" — `source/apps/panel/public/fonts` is ~1.0MB (10 TTFs at ~100KB each). The "~2.3MB today" total is closer to ~1.8MB.
3. `config/styles.json` is listed as a 24h-cached runtime asset, but `source/apps/panel/src/assets/styles/variables.styl:3` reads it at build time via stylus `json(...)`. Nothing fetches it at runtime, so its cache policy is inert.

On the 24h `sounds/*.mp3`: acceptable. Today they are `max-age=0` so a changed sound lands on next reload; after this it can lag up to a day. The failure mode is *the old alert sound plays*, not silence, and the risk is disclosed in the PR body.

zbyju:
All three corrections verified independently and applied to the body. No code change — the wiring is correct as reviewed.

**1. `public, max-age=0`, not "no `Cache-Control` at all"** — confirmed. `serve-static@1.15.0/index.js:64` does `opts.maxage = opts.maxage || opts.maxAge || 0`, and `send@0.18.0/index.js:862` emits `'public, max-age=' + Math.floor(this._maxage / 1000)` whenever no `Cache-Control` is already set. I ran the base wiring against this repo's `express@4.18.1` and got `public, max-age=0` on `/`, `/index.html`, `/assets/*` and `/sounds/*`.

This changes the performance argument more than the wording. `max-age=0` means assets stay in the cache and normally come back `304`, so today's cost is a round trip per file, not the bytes. The body previously said "either re-transferred or at minimum revalidated", which implied re-transfer was the common case. It is not — bytes only move on a cold cache. Rewritten to claim round trips, not bandwidth.

**2. Font sizes** — confirmed, your number is right. Measured from the checked-out tree:

| | bytes | |
|---|---|---|
| 10 Open Sans TTFs | 1,010,684 | ~1.0MB |
| `fonts/` incl. `opensans.css` | 1,012,138 | ~1.0MB |
| 5 `sounds/*.mp3` | 442,788 | ~0.44MB |
| all of `public/` | 1,457,325 | ~1.46MB |

Plus ~322KB of hashed JS/CSS, the total is ~1.78MB. "~1.8MB" is right; "~2MB of TTFs" and "~2.3MB total" were both wrong.

**3. `config/styles.json` is build-time only** — confirmed. `variables.styl:3` is `json('../../../public/config/styles.json')`, there is no runtime reference anywhere in `panel/src`, and a built `dist/assets/*.js` contains zero occurrences of the string. It is still copied into `dist/config/` and therefore still served with the 24h header, so the header is real but nothing ever requests it. Body now says exactly that instead of listing it as a runtime asset.

**One more the body got wrong, which the corrections above did not cover.** The MP3s are not fetched on page load at all. `panel/src/composables/useSounds.ts` constructs the `Howl` inside `playSound`, not at setup, so nothing touches `/sounds/*` until an alert actually fires. So:

- The "~5 alert MP3s" line in the page-load estimate was wrong. A repeat load is `index.html` + JS + CSS + `favicon.ico` + `opensans.css` + fonts.
- "roughly 20+ round trips" was overstated. The ceiling is ~15, and it is lower in practice because `opensans.css` is 10 separate `@font-face` rules and a browser only fetches the faces it renders. That also makes ~1.0MB of TTFs a ceiling rather than a typical load.
- The 24h policy on the sounds is still worth having, just for a different reason than the body claimed: `loadHowl` builds a fresh `Howl` on every `playSound`, so each alert re-requests its file. Today that is a conditional request per alert; after this it is a cache hit for a day.

Nothing here changes the verdict or the code. Body updated.

PR #79, opened 2026-09-10
fix: clear the thermal unit, not the engine unit, on a thermal fetch miss

zbyju:
`setRawThermalUnit` in the panel's units store cleared `engineUnit` instead of `thermalUnit` when the thermal unit returned no data. A copy-paste slip from `setRawEngineUnit` right above it.

The effect would be backwards: a thermal miss wipes the door state, clock and inspection timer from the panel, while the stale thermal readings stay on screen.

This is latent today. `getData` in `api/units.ts` rejects on failure instead of resolving `undefined`, so `panelLoop.ts` hits its catch and this branch never runs. The fix stops the bug from going live the moment that error path changes.


zbyju:
## Review

Low risk, nothing blocking. The change is correct and safe to merge.

- The fix is right: `setRawThermalUnit` cleared `engineUnit`, which would blank the door state, clock and inspection timer on a thermal miss. `source/apps/panel/src/pinia/unitsStore.ts:41` now clears `thermalUnit`.
- The latency claim checks out. `getData` returns `Promise.reject(undefined)` on any failure (`source/apps/panel/src/api/units.ts:26`), and on success it always returns an array from `body.data.split("|")`. It never resolves `undefined`, so `updateThermalUnit` at `source/apps/panel/src/logic/panel/panelLoop.ts:91` always lands in its `catch` and the branch never runs. Today the change is a no-op.
- No risk to what works now. Both units are already `undefined` at boot, and every consumer (`state.ts` via `thermalUnit?.data`, the four `Table*.vue` widgets) takes `Maybe` and uses optional chaining. Clearing `thermalUnit` cannot crash the panel or stop the loop.

Side note, not part of this PR: the same reject-not-resolve behaviour also makes the `data !== undefined` success check at `panelLoop.ts:93` dead.

zbyju:
Review read. No changes needed: the one-line fix and the PR description are both correct as they stand, and the review raised nothing blocking.

The `Promise.reject(undefined)` note in `api/units.ts` is a separate finding that is deferred, so it stays out of this PR.

PR #80, opened 2026-09-10
fix: mount the status route under the configured api prefix

zbyju:
The status route was the only route mounted from the raw `API_PREFIX` env var. Every other route uses `prefix`, which prefers `config.backend.url` from the configer service and only falls back to the env var.

The panel builds its status URL from the configer config, so on a machine where `.env` has no `API_PREFIX` the status route mounts at the literal path `undefined/status`. The panel's `getStatus()` then never succeeds and the panel sits on the "Načítám" boot screen forever.

These panels run unattended in hospitals with nobody on site, so there is no easy recovery once that happens.

The fix moves the `prefix` constant above the status route and mounts the status route with it. Fallback order is unchanged.


zbyju:
## Review

**Risk: low. Nothing blocking.**

- **No caller depends on the old path.** The only consumer of the backend `/status` is the panel (`source/apps/panel/src/api/units.ts:9`), and it builds the URL from `config.backend.url` (`source/apps/panel/src/api/base.ts:17`) — never from `API_PREFIX`. Nothing else in the repo calls it (no health check, no script; the Postman env uses the same `/api/v1`). So the new path matches the only caller, and the old one only worked when `.env` happened to agree with the config.
- **No ordering or shadowing problem.** `prefix` is the only binding of that name in `main()`, it is declared after `config` is assigned, and it is now declared before every use. Fallback order for the other routes is unchanged.
- **The claim checks out.** With `API_PREFIX` unset, Express 4 registers the literal path `undefined/status`, which no request can match. `initializeGlobal` only clears the boot interval on a successful status (`panelLoop.ts:205`), so the state stays `Trying` and the panel shows "Načítám" forever (`AppState.vue:8`).
- **Minor, not a defect:** configer merges `base.json` into `main.json` (`services/db/main.ts:26`), so `backend.url` is always set. The env fallback is therefore dead in practice, and `API_PREFIX` in `backend/.env` no longer affects anything. This also matches what configer already does (`configer/src/index.ts:30`).

zbyju:
Read the review — no changes needed. Three points confirm the fix and its reasoning; the fourth (configer merging `base.json` means `backend.url` is always set, so the env fallback is unreachable in practice) is an observation, not a defect. Keeping the `API_PREFIX` fallback on purpose: it costs nothing and guards the case where the config is missing.

PR #81, opened 2026-09-10
fix: cache the configer db init promise so it runs once

zbyju:
`DbFactory` cached the resolved db object, not the promise. Between the `if` check and the `await` finishing, a second concurrent call saw an empty cache and started its own init.

Now the promise itself is cached, so concurrent callers await the same in-flight init. Call sites are unchanged — all four already `await` the getter. The types are fixed too: `MainDb` / `VersionDb` are already promise types, so the old `async ...: MainDb` annotations and the `any` fields are gone.

## Scope: defensive, not urgent

`index.ts:15` awaits `getMainDb()` before `app.listen` (`index.ts:43`), so `mainDb` is always warm before the first HTTP request arrives. No request can start a second `mainConfig()`. The "two interleaved writes to `main.json`" scenario is **not reachable** on the current boot path — an earlier version of this description overstated it.

The only getter that can really race is `getVersionDb`, which is never warmed at boot. `versionConfig()` only reads `versions.json`, so a duplicate init there wastes a file read and nothing more.

So this is a correctness cleanup that keeps a future boot-order change from turning into a real problem, not a fix for a live bug.

## Caveat

A rejected init promise now stays cached forever — every later call gets the same rejection, with no retry. This is harmless today only because both failure paths kill the process: `main()` at `index.ts:54` has no `.catch()`, and Express 4 does not catch async handler rejections. If someone later adds an error handler, a failed init would turn into a permanent outage. Left as is on purpose; no retry logic in this PR.


zbyju:
## Review

Low risk. Nothing blocking.

Checked all four call sites (`index.ts:15`, `configRoute.ts:8,13,24`) — every one already awaits, so dropping `async` changes nothing for them. `mainConfig` and `versionConfig` are both `async`, so they cannot throw synchronously, and the getters contain no other throwing statement, so the sync-throw path does not exist. Types are fine: `MainDb` is `Promise<{data,update}>`, the field is `MainDb | undefined`, and truthiness narrowing gives `MainDb` on both branches, so the non-async return matches.

The write claim holds: `main.ts:28` calls `db.write()` during init, and `main.json` is not in the repo, so first boot creates it.

Two notes, neither blocking:

- The trigger is overstated. `index.ts:15` awaits `getMainDb()` before `app.listen` (`index.ts:43`), so `mainDb` is always warm before the first request. No request can start a second `mainConfig()`. The only getter that can really race is `getVersionDb`, never warmed at boot — and `version.ts` only reads. The fix is correct, but defensive, not a live corruption path.

- Cached rejections are harmless only because both failure paths kill the process: `main()` at `index.ts:54` has no `.catch()`, and Express 4 does not catch async handler rejections. If someone later adds an error handler, a failed init would stay cached and every later request would fail.

zbyju:
Both notes addressed in the description, no code change.

Verified the first one against the code: `index.ts:15` awaits `getMainDb()` and `app.listen` is at `index.ts:43`, so the main cache is warm before any request. `configRoute.ts` cannot race it. `versionConfig()` (`version.ts:8-19`) only does `db.read()` and returns a `data()` getter, so the one getter that can race is read-only. The description now says the interleaved-write scenario is not reachable and calls the fix defensive.

The rejection-caching caveat is recorded as its own section, including why it is harmless today (`main()` has no `.catch()` at `index.ts:54`, Express 4 does not catch async handler rejections) and what breaks if an error handler is added later. No retry or cache-invalidation code — out of scope.

The diff itself is unchanged: the caching, the types, and the four call sites all check out.

PR #82, opened 2026-09-10
fix: always poll the units at the configured request delay

zbyju:
`startPanelLoop` read the panel state once at mount and then picked the interval period with a ternary, so the "poll faster" branch was never adaptive.

On a fresh boot the default panel state has no message, so the `delay / 2` branch never runs at all — it is dead on the path that actually matters.

`MainView` starts the loop on mount and stops it on unmount, and it is remounted every time the user navigates back to `/`. If an alert message is showing at that moment, the loop locks to `delay / 2` — 1s instead of 2s — for the rest of the day, doubling the request load on the engine and thermal units.

That is not a change to make by accident based on which screen the user happened to be on, so this PR just removes the accidental version and always uses the configured delay.

A genuinely adaptive interval is a real behaviour change with a hardware-load cost. It is deliberately left for a separate PR with the owner's sign-off.


zbyju:
## Review

**Risk: low. Nothing blocking.** Traced the restart path end to end; the change cannot cause a machine restart.

- **The restart heartbeat does not go through the panel loop.** `modules.onIncomingData()` is called from exactly one place, `GET /restart/refresh` (`source/apps/backend/src/routes/restartRoute.ts:9`), and the panel hits that from its own fixed 5000 ms interval in `source/apps/panel/src/App.vue:18-24`, outside `MainView` and independent of `requestDelay`. The loop's `updateWatchdogEngine` calls `PUT /engine/watchdog` (`source/apps/backend/src/routes/engineRoute.ts:29`), which only pets the engine unit and never touches `lastRequest`. So `RESTART_INTERVAL` (20 s) and `RESTART_ERROR_THRESHOLD` (9) in `source/apps/backend/src/modules/restart.ts:25-26` keep their ~180 s margin against an unchanged 5 s heartbeat. Zero effect from this diff.

- **The engine-unit watchdog pet slows to 2 s only in the remount case, and 2 s is already the normal rate.** Every install that has not navigated away from `/` since boot has always run at the configured 2000 ms (`configs/base.json:23`). This change makes the remount case match that, not exceed it. No new worst case.

- **PR claim verified.** Store and default both start `message: undefined` (`panelStateStore.ts:8`, `panelState.default.ts:5`), and `MainView` starts the loop in `onBeforeMount` before any tick, so `delay / 2` was unreachable on a fresh boot.

- **No leftovers.** `this.panelState` is still used at `panelLoop.ts:113`; all imports still used.

zbyju:
Review read. Nothing actionable: all four points confirm the change rather than request one, and the PR body makes no claim about the restart heartbeat, so no code or description change. Adaptive polling stays deferred to a separate PR.

PR #83, opened 2026-09-10
perf: drop the fixed sleeps that delay boot

zbyju:
Two fixed sleeps made the system slower to boot for no reason. Both are gone, and the boot screen now has a real minimum time so it does not flash.

## Backend: no 2 s sleep before fetching the config

`fetchConfig` slept 2 s before it even tried to reach configer. `main()` in `index.ts` already retries in a loop with its own 5 s gap, so this sleep only added 2 s to the first attempt and 2 s to every retry. Removed the sleep and the now-unused `wait` import. The retry loop is untouched.

**Caveat — cold start where configer is not listening yet.** The 2 s sleep used to double as a head start for configer. Without it, the first fetch now fails in milliseconds instead of succeeding at t+2 s, so the config lands on the 5 s retry instead — roughly 3 s later than before in that one case. Every other case is 2 s faster. This is accepted, not worked around: `startup` launches configer before the backend, so in the deployed path configer is already up and the fast path is the one that runs. No retry, backoff, or replacement sleep was added.

## Panel: first health check fires immediately

`initializeGlobal` used `setInterval`, which never runs its body straight away. The panel therefore waited a full 5 s before its first config/backend attempt, even when both services were already up. The tick body is now a local function that runs once immediately and then keeps polling every 5 s.

Behaviour is unchanged otherwise: the config is still fetched once, a config failure still stops the polling and sets the config error, backend success still stops the polling, backend failure still keeps polling. A `stop()` helper clears the interval, and tolerates the handle not being set yet.

## Boot screen stays up for at least 1 s

With the two sleeps gone the panel can be ready almost instantly, and the "Načítám" screen would flash by. The old code delayed the switch by 500 ms measured from completion, which gives no real guarantee.

Now the store records a start timestamp and delays the switch by `max(0, 1000 - elapsed)`, so the screen is visible for at least 1000 ms measured from app start. If more than a second has already passed, it switches with no extra delay. `performance.now()` is used instead of `Date.now()` because these panels run unattended for months and an NTP step backwards must not stall the boot. A flag stops repeated calls from stacking timers.


zbyju:
## Review

Low risk. Nothing blocking. The sharp hazards are all handled correctly.

**Panel `initializeGlobal`** (`source/apps/panel/src/logic/panel/panelLoop.ts:185-231`)

- No TDZ / ReferenceError. `attempt()` suspends at its first `await` (or at the `.then` registration), so `stop()` can never run before `interval` is assigned on line 229. `interval` is a `let` declared on line 188, so reading it is safe in any case. This is the failure mode that would have killed boot; it is not reachable.
- All four paths kept: config fetched once via `configLoaded`; config failure still stops polling and sets the config error; backend success still stops polling and sets success; backend failure still keeps polling.
- No new overlap. Ticks are still 5 s apart, and the immediate attempt only replaces the tick that used to be at t+5 s. Attempts could already overlap when `getConfig` hangs (it has no timeout, `panelLoop.ts:144`) — pre-existing, unchanged.
- Dead guard: `if (!stopped)` on line 228 can never be false, for the reason above. Harmless.

**Boot floor** (`source/apps/panel/src/pinia/appStateStore.ts:12-59`)

- Correct. `performance.now()` throughout, `startedAt` set when the store is first created (AppManager constructor, in `App.vue` setup), `okScheduled` blocks stacked timers, and every path that reached `AppState.Ok` before still reaches it.

**Backend** (`source/apps/backend/src/fetch/fetchConfig.ts`)

- `wait` is still imported and used in `index.ts:14`; only the unused import was dropped. Retry cadence goes 7 s -> 5 s against localhost, which is not hammering.
- One caveat: on a cold start where configer is not listening yet, the first call now fails in milliseconds instead of succeeding at t+2 s, so the config can land ~3 s later than before. `startup` starts configer first and `start:main` runs `pnpm install` before launching, so configer should be up. Low.

zbyju:
Addressed the two actionable points.

**Dead guard — removed** (d5e149e). Verified before touching it: `stopped` is written only inside `stop()`, read only at the `if (!stopped)`, and `stop()` is reachable only from `.then`/`.catch` microtasks. Those cannot run while `initializeGlobal`'s synchronous block is on the stack, and there is no `await` between `attempt()` and the check, so `stopped` was always `false` there. `setInterval` is now called unconditionally. The `interval !== undefined` check inside `stop()` stays — `interval` is `Maybe<...>`, so it is a type guard as much as a runtime one.

**Cold-start caveat — documented, not fixed.** Added to the PR description: without the 2 s sleep, a cold start where configer is not yet listening fails in milliseconds and falls to the 5 s retry, so ~3 s slower in that one case, 2 s faster in every other. No retry, backoff, or replacement sleep added — `startup` launches configer first, so the deployed path takes the fast route.

**Not acted on:** the four confirmation bullets, and `getConfig` having no timeout — that one is pre-existing on `main` and out of scope here.

No `pnpm`/lint run — no `node_modules` in this worktree and the repo pins pnpm@7.5.0. The change is a pure deletion, so nothing new to type-check.

zbyju:
Correction on timing: this PR was squash-merged (9bba599) before my cleanup commit was pushed, so the squash contains only 7ec1781. The dead-guard removal (d5e149e) is on `origin/perf/remove-fixed-boot-delays` but is **not** in `main` — `if (!stopped)` is still there on `main` today.

Nothing is lost; the commit is safe on the branch. It needs a small follow-up PR to land. Not opening one without a go-ahead.

The PR description update did apply.

PR #85, opened 2026-09-11
feat: make writing the config safe

zbyju:
Phase **P0** of [docs/plans/config-ui.md](docs/plans/config-ui.md): make writing the
configer config safe before any UI is built on top of `PUT /config/main`.

## Why

Every deployed box keeps its own `configs/main.json`. Today a `PUT` can drop the unit
IPs past the validator and replaces the whole file. The write itself already went
through steno (lowdb 3's `JSONFile`), which writes a temp file and renames it, but it
never calls `fsync`, so after a power cut the rename can land before the data and leave
an empty or truncated file. There is also no backup, and `JSON.parse` in
`JSONFile.read()` throws, so a corrupt file stops configer from starting. Neither the
backend nor the panel starts without a readable config. That is a dead babybox that
needs someone on site.

## What changed

**Validator** (`src/types/main.types.ts`)

- One function, `validateMainConfig(config)`, returns `{ path, msg }` per problem.
  The old `isInstanceOfMainConfig` boolean wrapper is deleted; it had no caller left.
- `MainConfigUnits` gained `engine` and `thermal`, and both are checked, including
  `units.engine.ip` / `units.thermal.ip`.
- `startup` is checked.
- `app.refreshRequestLimit` must be an integer when present.
- `camera.cameraType` is a union of the names the panel understands: `dahua`,
  `hikvision`, `avtech`, `avm`, `vivotek`. The check ignores letter case, because
  deployed files hold `DAHUA` as well as `dahua`. `pc.os` stays an exact match, the
  backend compares it with `=== "ubuntu"`.
- The per-section guards are gone. One shape check, not eight.

**Write path** (`src/services/db/main.ts`)

- `update(body)` merges the body over `base.json`, validates, then writes. Same order
  as boot, so a partial body gets defaults instead of missing keys. It returns
  `saved`, `invalid` (with field errors) or `write-failed`; nothing is written when
  invalid, and memory keeps the old config when the write fails.
- A non-object body is rejected up front. `lodash.merge` would otherwise spread a
  string over the defaults.
- An empty body is rejected too. `express.json()` leaves `req.body` as `{}` when the
  Content-Type is not JSON, and an empty merge over `base.json` would reset the box.
- The write is `main.json.tmp` -> `fsync` -> `rename` over `main.json` -> `fsync` of
  the directory. The current file is copied to `main.json.bak` first, and only when
  it is a JSON object, so a corrupt file can never overwrite a good backup.
- Boot: missing `main.json` starts from `base.json` (unchanged). A `main.json` that
  does not parse falls back to `main.json.bak` with a `console.warn`; if that fails
  too it starts from `base.json` with a `console.error`. It never throws. The
  unreadable file is copied to `main.json.corrupt` first, so a hand-edit with a typo
  is still there to recover from.
- lowdb is dropped for `main.json` and replaced with `node:fs`, because steno gives no
  hook for the `fsync` or the backup. `versions.json` keeps lowdb. Rationale in
  [docs/decisions.md](docs/decisions.md).
- The config directory is a parameter defaulting to the current path, so tests point
  it at a temp dir.

**Route** (`src/routes/configRoute.ts`)

- `PUT /main` returns `400 { msg, errors: [{ path, msg }] }`, or `500 { msg }` when
  the disk write fails. The body is no longer echoed back.

**Tests** — vitest tests next to the code (`main.test.ts`, `main.types.test.ts`,
`configRoute.test.ts`), plus a "Configer tests" step in CI.

**Docs**

- `docs/mottos.md` — how we decide when two options both work. Every agent reads it
  before it plans, implements or reviews.
- `docs/learnings.md` — things that cost us time once, one line per lesson (for
  example: never run the machine's pnpm here).
- `docs/plans/config-ui-agent-workflow.md` — the implement -> review -> verify ->
  post -> address pipeline that builds each phase of the plan, with the PR as the
  record.

## Deliberately left out

- `PATCH /config/main` (partial update over the current file) — P2.
- Auth on the write endpoint — P5. It is still unauthenticated, and still unused.
- zod and the shared schema package — P1, after the zod decision.
- Config history and rollback beyond the one `.bak` — open question in the plan.
- The panel and the backend are untouched.

## How to test

```bash
cd source
npx pnpm@7.5.0 install --frozen-lockfile
npx pnpm@7.5.0 -F babybox-panel-configer exec vitest run
npx pnpm@7.5.0 -F babybox-panel-configer build
```

By hand: start configer, `PUT` a body with `units.engine.ip` set to a number, and check
the 400 lists `units.engine.ip`. `PUT` `{"babybox":{"name":"X"}}` and check `main.json`
still holds every other key and `main.json.bak` holds the previous file. Corrupt
`main.json`, restart, and check it boots from the backup and warns.

## Upgrade safety

- **`main.json` shape unchanged.** The file is still written with
  `JSON.stringify(config, null, 2)`, exactly what lowdb wrote. Existing files are read
  and merged over `base.json` the same way as before.
- **Lockfile format unchanged.** Still `lockfileVersion: 5.4`, produced by
  `pnpm@7.5.0`, and `--frozen-lockfile` passes. The only new package is vitest,
  pinned to `^0.9.3` — the version the panel already has, so no new download. The diff
  also dedupes `function-bind` 1.1.1 into 1.1.2, which was already in the lockfile;
  worth a look during review.
- **Node 18 only.** `openSync`, `fsyncSync`, `renameSync`, `copyFileSync`, `mkdtempSync`
  are all long-standing APIs.
- **Build prints nothing on stderr,** which the startup app treats as a failed update.
- **Tests stay out of `dist`.** `tsconfig.json` excludes `src/**/*.test.ts`, so pm2
  still starts `dist/index.js` and nothing else.


zbyju (commented):
Automated review: 10 findings, posted as inline comments below. Each thread will be answered and resolved by the address step.

source/apps/configer/src/services/db/main.ts
```diff
@@ -1,40 +1,110 @@
...
+      fsyncSync(handle);
+    } finally {
+      closeSync(handle);
+    }
+    renameSync(tempFile, mainFile);
+  }
+
+  // Boot is not validated on purpose; an odd stored value must not stop the box.
+  let data = merge(freshBase(), loadStored(mainFile, backupFile)) as MainConfig;
+  write(data);
```
zbyju: [blocker] Boot destroys an unparseable main.json with no copy left.

When main.json does not parse, `loadStored` returns the .bak (or nothing) and `write(data)` renames the fallback over main.json. The backup step is skipped on purpose (`parseFile(mainFile)` is undefined), so the maintainer's original file is gone.

Reproduced with the built dist: a hand-edited main.json with one trailing comma (`{"babybox":{"name":"Ostrava"},"units":{"engine":{"ip":"10.9.9.9"}},}`) and no usable .bak boots with base.json defaults ("Nenastaveno", 10.1.1.5), main.json is overwritten, and the only trace is one console.warn in the pm2 log. With a good .bak present the box keeps the previous config, but the edit is still destroyed. A UTF-8 BOM from Notepad does the same, because `JSON.parse` rejects it. main.json is edited by hand on every box until the UI ships, so a typo plus a restart is a normal event. Before this PR lowdb threw on the same file, so configer refused to start and the file stayed on disk for someone to fix.

Fix, one line: in `loadStored`, when the file exists but does not parse, `copyFileSync(mainFile, join(configDir, "main.json.corrupt"))` before falling back, and add `main.json.corrupt` next to `.bak` and `.tmp` in the configer .gitignore. The fallback still keeps the box up, and the edit is still there to recover from. A test that the corrupt copy exists after boot would fit next to "falls back to base.json when both files are corrupt".

source/apps/configer/src/services/db/main.ts
```diff
@@ -1,40 +1,110 @@
...
+
+  /*
+   * A full replace. The body is filled in from base.json first, so a client that
+   * leaves a key out gets the default, never a missing key.
+   */
+  async function update(body: unknown): Promise<UpdateResult> {
+    // lodash.merge would spread a string or an array over the defaults.
+    if (typeof body !== "object" || body === null || Array.isArray(body)) {
```
zbyju: [should] `update({})` passes the object check, merges nothing over base.json, validates, and writes. Every field goes back to defaults: unit IPs, camera credentials, app password, babybox name. Checked against the built code: it returns `ok: true` and main.json holds base.json afterwards.

This is easy to hit by accident. `express.json()` sets `req.body = {}` for any request whose Content-Type is not JSON (body-parser `lib/types/json.js` runs `req.body = req.body || {}` before the type check). So `curl -X PUT -d @main.json` without `-H 'Content-Type: application/json'` sends form-urlencoded, the parser skips it, the route returns 200, and the box is on defaults. Reproduced with the same `express.json()` setup as `src/index.ts`: form-urlencoded, text/plain, and no header all wrote defaults with a 200.

main.json.bak holds the previous config only until the next configer restart. Boot calls `write(data)` unconditionally, which copies the now-default main.json over the .bak. One mistyped command plus one restart and the last good config is gone.

On `main` this request got a 400, because `isInstanceOfMainConfig({})` failed with no merge. The merge-over-base in this PR opened the hole.

Nothing needs "reset everything" through PUT; a reset would send the full default body. Fix, one line at the top of `update` after the object check:

```ts
if (Object.keys(body).length === 0) return { ok: false, errors: [{ path: "", msg: "body is empty" }] };
```

That turns the missing-header mistake into a 400 instead of a silent wipe. Worth a test next to the "non-object body" one.

source/apps/configer/src/services/db/main.ts
```diff
@@ -1,40 +1,110 @@
...
+      return { ok: false, errors: [{ path: "", msg: "must be an object" }] };
+    }
+
+    const merged = merge(freshBase(), body);
+    const errors = validateMainConfig(merged);
+    if (errors.length > 0) return { ok: false, errors };
+
+    data = merged as MainConfig;
```
zbyju: [should] `data = merged` runs before `write(data)`. If `write` throws (disk full, or EPERM on a Windows box when an antivirus scan holds `main.json` during the rename), memory already holds the new config while disk still has the old one.

Today nothing observes that: the rejection is not caught (Express 4, no error middleware in `index.ts`), Node 18 exits on the unhandled rejection, pm2 restarts configer and boot reloads from disk. I checked this with a small Express 4 test; no request is served between the throw and the exit. So the crash hides the mismatch, and that crash is pre-existing (lowdb's write could throw too), so I am not asking for a try/catch here.

The order is still wrong on its own. The moment a try/catch or error middleware lands (a 500 on write failure is the natural P2 step), `GET /main` starts returning a config that was never saved, and a later `PUT` backs up the wrong "previous" file. The fix adds no code: call `write(merged as MainConfig)` first, then `data = merged as MainConfig`, then return. Memory then never runs ahead of disk.

docs/decisions.md
```diff
@@ -0,0 +1,100 @@
...
+- Decision: do P0 with the existing hand-written guards. Fix the holes by hand.
+  Decide zod in P1.
+- Why: P0 is useful on its own and must not wait on a dependency decision. It also
+  stays small enough to review in one PR.
+- Gave up: the guards stay duplicated for now.
+- Where: this file, PR "feat: make writing the config safe".
+
+## 2026-09-11 — Config file is written atomically, with one backup
+
+- Context: lowdb truncates then rewrites `main.json`. A power cut mid-write leaves a
```
zbyju: [should] The premise here is wrong. lowdb 3.0.0's `JSONFile` writes through steno 2.1.0 (`source/node_modules/.pnpm/steno@2.1.0/node_modules/steno/lib/index.js`). steno writes to `.main.json.tmp` and then renames over `main.json`. It does not truncate and rewrite in place. The old write was already a temp file plus rename.

What the old path did not have: an `fsync` before the rename (without it the rename can land before the data, so after a power cut `main.json` can be empty or truncated on some filesystems), a backup copy, and a boot that survives a corrupt file (`JSON.parse` in `JSONFile.read()` threw, so configer did not start). Those are the real reasons for this PR, and the `fsync`, the `.bak` and the fallback are the real changes.

As written, the next person reading this log will think the rename was the fix and may drop the `fsync` when they touch this again.

Fix: rewrite the Context line of this entry and of "main.json is read and written with node:fs, not lowdb" to say lowdb already wrote a temp file and renamed it, but did not `fsync`, kept no backup, and threw on a corrupt file. Risk 3 in `docs/plans/config-ui.md` (line 55) and the "Why" paragraph of the PR body make the same claim and need the same correction.

docs/decisions.md
```diff
@@ -0,0 +1,100 @@
...
+- Decision: one Opus agent implements a step and opens a PR. Review agents post
+  verified findings as inline PR comments. An Opus agent addresses each thread,
+  replies and resolves it. See
+  [config-ui-agent-workflow.md](plans/config-ui-agent-workflow.md).
+- Why: the PR is the record. Anyone can read what was found and what was done.
+- Gave up: speed; a review round adds an hour.
+
+## 2026-09-11 — main.json is read and written with node:fs, not lowdb
+
+- Context: P0 needs an atomic write with a backup. lowdb's `JSONFile` truncates and
```
zbyju: [nit] The reason recorded for dropping lowdb is not what the code on main did. Lines 26 and 67 say lowdb truncates then rewrites `main.json` in place. lowdb 3.0.0's `JSONFile` writes through steno 2.1.0, and `source/node_modules/.pnpm/steno@2.1.0/node_modules/steno/lib/index.js` (lines 18, 59, 60) writes to `.main.json.tmp` and then renames it over the target. So a power cut mid-write could not leave a half-written `main.json` on main either.

What was missing on main: no `fsync` before the rename (so a rename-only write can still end up as an empty file after a power cut on ext4), no `main.json.bak`, and no boot fallback. That is what this PR adds, and it is still a good reason to own the write.

Please reword both entries, and the matching sentence at `docs/plans/config-ui.md` line 55, to say that. Otherwise the next person may build a decision on a wrong fact, for example treating `versions.json` on lowdb as at risk of in-place truncation. No code change needed.

source/apps/configer/src/types/main.types.ts
```diff
@@ -9,175 +33,176 @@ export interface MainConfig {
...
+    checkString(errors, app.password, "app.password");
+    if (app.refreshRequestLimit !== undefined) {
+      checkInteger(errors, app.refreshRequestLimit, "app.refreshRequestLimit");
+    }
+  }
+
+  return errors;
+}
+
+export function isInstanceOfMainConfig(obj: unknown): obj is MainConfig {
```
zbyju: [nit] `isInstanceOfMainConfig` has no caller. The only reference left in `source/apps` is this definition. The one caller on `main` was `configRoute.ts`, and this PR changed it to call `main.update()`, which uses `validateMainConfig` directly. The tests import `validateMainConfig` only, and configer is not imported by any other app.

The PR body says the wrapper stays "so callers are unchanged", but there are none, so it is dead code. Delete lines 206-208. If a later phase needs a boolean, it is one line to add back. `tsc` cannot break, since nothing imports the symbol.

source/apps/configer/src/services/db/main.ts
```diff
@@ -1,40 +1,110 @@
...
-
-  async function update(c2: MainConfig): Promise<MainConfig> {
-    db.data = c2;
-    await db.write();
-    return db.data;
+export type UpdateResult =
+  | { ok: true; config: MainConfig }
+  | { ok: false; errors: ConfigError[] };
+
+export const defaultConfigDir = join(
```
zbyju: [nit] `defaultConfigDir` is exported but nothing imports it. Its only use is the default value of the `configDir` parameter in this file. `factory.ts` calls `mainConfig()` with no argument and the tests pass their own temp dir.

Either drop the `export` and keep it module-local, or give it a real consumer: `main.test.ts` builds the same `../../../configs` path by hand as `repoBase`, which could be `join(defaultConfigDir, "base.json")`. One of the two, not both.

source/apps/configer/src/services/db/main.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    const db = await mainConfig(configDir);
+
+    expect(db.data()).toEqual(base());
+    expect(error).toHaveBeenCalled();
+  });
+});
+
+describe("update", () => {
+  it("fills a partial body from base.json", async () => {
```
zbyju: [should] This test cannot tell "merge the body over base.json" from "merge the body over the current stored config". It starts with main.json equal to base.json, so both give the same file. The other update tests change the same key twice, so they cannot tell either.

I checked: with main.ts:97 changed to `merge({}, data, body)` the whole suite stays green (23/23). That is the exact behaviour decisions.md records ("Writes merge over base.json, same as boot") and the sharp edge of PUT: a key set by an earlier PUT goes back to its default when a later partial PUT leaves it out. Nothing pins it. Someone adding PATCH in P2 could switch the merge base and no test would fail.

Fix, one line, no src change: before `db.update({ babybox: { name: "Brno" } })` add `await db.update({ camera: { ip: "10.1.1.99" } });`. The existing `toEqual({ ...base(), babybox: { name: "Brno" } })` then already asserts camera.ip is back to the base.json default (10.1.1.7). With that line the test passes on the current code and fails on the merge-over-stored variant.

source/apps/configer/src/services/db/main.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    const result = await db.update({ babybox: { name: "Brno" } });
+
+    expect(result.ok).toBe(true);
+    expect(readJson("main.json")).toEqual({
+      ...base(),
+      babybox: { name: "Brno" },
+    });
+  });
+
+  it("rejects an invalid body and writes nothing", async () => {
```
zbyju: [should] The non-object guard in update() (main.ts:93) has no test. It is the only thing that stops an array body. express.json() in strict mode accepts `[]` and `["x"]`, so both reach update(). Without the guard, `merge(base, [])` returns base.json unchanged, validation passes, and the whole config is silently reset to defaults. `["x"]` writes a config with a `"0"` key, which the validator does not catch. The guard also looks redundant next to the root checkObject in validateMainConfig, so a tidy-up could delete it and nothing would go red. The `validateMainConfig("nope")` test does not cover this; it never goes through update() or the merge.

Fix: add one test next to "rejects an invalid body and writes nothing":

```ts
it("rejects a non-object body and writes nothing", async () => {
  const db = await mainConfig(configDir);
  const before = readFileSync(file("main.json"), "utf-8");

  const result = await db.update([]);

  expect(result).toEqual({
    ok: false,
    errors: [{ path: "", msg: "must be an object" }],
  });
  expect(readFileSync(file("main.json"), "utf-8")).toBe(before);
});
```

source/apps/configer/src/types/main.types.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  for (const cameraType of cameraTypes) {
+    it(`accepts the camera type ${cameraType}`, () => {
+      const config = baseConfig();
+      config.camera = { ...(config.camera as object), cameraType };
+
+      expect(validateMainConfig(config)).toEqual([]);
+    });
+  }
+
+  it("accepts a camera type in upper case", () => {
```
zbyju: [should] pc.os has no test. The check at main.types.ts:193 is exact-case on purpose (decisions.md: the backend does `pc.os === "ubuntu"` in modules/restart.ts), but nothing pins that. The test above proves camera.cameraType accepts "DAHUA"; the mirror case for pc.os is missing, so a later "make every enum check case-insensitive" cleanup would pass this suite. Then "Ubuntu" gets written to main.json, the backend's ubuntu branch does not match, and the box runs the Windows shutdown command.

Fix: add two tests here, no src change.
- `rejects pc.os in a different case`: set pc.os to "Ubuntu", expect an error at path "pc.os".
- `rejects an unknown pc.os`: set pc.os to "linux", expect an error at path "pc.os".

source/apps/configer/src/services/db/main.ts
zbyju: Fixed in "fix: keep an unreadable main.json as main.json.corrupt". loadStored now copies the file to main.json.corrupt before it falls back, the warn line names the copy, and main.json.corrupt is in the configer .gitignore next to .bak and .tmp. Added the test "keeps the unreadable main.json as main.json.corrupt".

source/apps/configer/src/services/db/main.ts
zbyju: Fixed in "fix: reject an empty body instead of resetting the config". update() now returns { path: "", msg: "must not be empty" } for a body with no keys, so the route answers 400. You are right that the merge-over-base opened this; nothing needs a reset through an empty body. Added a test that an empty body leaves main.json untouched.

source/apps/configer/src/services/db/main.ts
zbyju: Fixed in "fix: write the config to disk before updating memory". update() now calls write(merged) first and assigns data afterwards, so memory never runs ahead of disk. Agreed on the reasoning: the crash only hides it today, and it breaks the moment a try/catch or error middleware lands.

docs/decisions.md
zbyju: You are right, I checked steno 2.1.0 and lowdb 3.0.0 TextFile myself: the old path already wrote a temp file and renamed it. Fixed in "docs: correct the lowdb write premise and record the review outcomes" - both decisions entries and risk 3 in config-ui.md now say the rename was already there and the missing pieces were the fsync, the backup and a boot that survives a corrupt file. I will correct the PR body too.

docs/decisions.md
zbyju: Same fix as the thread above, in "docs: correct the lowdb write premise and record the review outcomes". Both entries and config-ui.md line 55 now say lowdb wrote through steno with a temp file and a rename, and that the fsync, the .bak and the boot fallback are what this PR adds. Also added a learning so the next person checks the dependency source before writing down why we replaced it.

source/apps/configer/src/types/main.types.ts
zbyju: Agreed, it has no caller. Deleted in "refactor: drop the dead config guard and reuse defaultConfigDir". The PR body line about keeping callers unchanged was left over from an earlier draft.

source/apps/configer/src/services/db/main.ts
zbyju: Fixed in "refactor: drop the dead config guard and reuse defaultConfigDir". Kept the export and gave it the real consumer: main.test.ts now builds repoBase as join(defaultConfigDir, "base.json") instead of resolving the same path by hand, so the path is written once.

source/apps/configer/src/services/db/main.test.ts
zbyju: Fixed in "test: pin the merge base, the non-object body and the pc.os case". The test now sets camera.ip first and is renamed "fills a partial body from base.json, not from the stored config". I checked your variant: with merge({}, data, body) the suite goes red on exactly that test and stays green everywhere else.

source/apps/configer/src/services/db/main.test.ts
zbyju: Fixed in "test: pin the merge base, the non-object body and the pc.os case". Added "rejects a non-object body and writes nothing" with an array body, as you wrote it. The guard is not redundant next to the root checkObject, because merge(base, []) returns a valid base.json and validation passes.

source/apps/configer/src/types/main.types.test.ts
zbyju: Fixed in "test: pin the merge base, the non-object body and the pc.os case". Added "rejects pc.os in a different case" ("Ubuntu") and "rejects an unknown pc.os" ("linux"), both asserting the error path is pc.os. That pins the exact-case check against a later case-insensitive cleanup.

source/apps/configer/src/services/db/main.ts
```diff
@@ -1,40 +1,135 @@
...
+    }
+    renameSync(tempFile, mainFile);
+  }
+
+  // Boot is not validated on purpose; an odd stored value must not stop the box.
+  let data = merge(
+    freshBase(),
+    loadStored(mainFile, backupFile, corruptFile)
+  ) as MainConfig;
+  write(data);
```
zbyju: `rv-doubt` 🟡 **Boot rewrites the config, so the backup can never hold an older config.**

`write(data)` runs on every boot. The backup copy at line 84 sits inside `write()`, so every boot copies the current `main.json` over `main.json.bak`. The backup holds the config as of the last boot, not the config before the last change. It protects against one failure only: a `main.json` that does not parse. It gives nothing against a write that is valid JSON and wrong, which is the case the UI will produce.

**Scenario.** An operator sets `units.engine.ip` to a wrong address. It is a string, so the validator accepts it. `write()` puts the good config in `.bak` and the wrong one in `main.json`. The box reboots before anyone looks (power cut, pm2 restart, unattended update). Boot parses `main.json`, copies it over `.bak`, writes it back. The good config is gone from both files.

**Evidence.** No process outside configer reads `main.json` from disk. The backend uses `fetchConfig()` (`source/apps/backend/src/index.ts:61`). The startup app only runs `pm2 start configer`. `main.test.ts:166` does two `update()` calls in one process; no test boots twice, so the reset is invisible in the suite. The two rules in `docs/decisions.md` ("a corrupt main.json never overwrites .bak", "an unreadable main.json is kept as .corrupt") exist only because boot writes.

**Alternative.** Make boot read-only: merge `base.json` and the stored file in memory, write only in `update()`. Then `.bak` is "the config before the last PUT", a real one-step rollback. `main.json.corrupt`, its gitignore entry, and the `parseFile` guard in `write()` are not needed. Tradeoff: after a boot `main.json` shows only the keys someone typed, not the full merged shape. `GET /config/main` and `base.json` still show the full shape. This is outside the P0 checklist.

**Decision needed.** Confirm whether the boot write has a consumer outside this repo (deploy script, image build, support runbook). If none: delete `write(data)` here, drop `corruptFile` and the guard at line 84, adjust four boot tests. If one exists: keep the boot write and give the backup its own trigger, not "every write". Confidence medium.

source/apps/configer/src/routes/configRoute.ts
```diff
@@ -15,13 +14,12 @@ router.get(["/version", "/versions"], async (req: Request, res: Response) => {
 });
 
 router.put("/main", async (req: Request, res: Response) => {
-  const c = req.body;
-  if (!isInstanceOfMainConfig(c)) {
+  const main = await DbFactory.getMainDb();
+  const result = await main.update(req.body);
```
zbyju: `rv-logic` · `rv-types` 🟡 **A failed write kills the process.**

Nothing catches a throw from `write()`, and `UpdateResult` (`main.ts:22`) has no write-failure case. Express 4 does not forward a rejected async handler to error middleware, so the client never gets a reply and Node 18 exits on the unhandled rejection. A full SD card or a read-only remount is the normal failure on a box.

1. The disk is full.
2. PUT passes validation, `writeFileSync` throws ENOSPC inside `update()`.
3. The handler's promise rejects, Express 4 ignores it.
4. Configer exits, the client waits until timeout.

Fix: wrap `write(merged)` in `update()` in try/catch and return `{ ok: false, errors }` (or a third union case), or add an Express error handler and answer 500.

source/apps/configer/src/services/db/main.ts
```diff
@@ -1,40 +1,135 @@
...
+    if (parseFile(mainFile) !== undefined) copyFileSync(mainFile, backupFile);
+
+    const handle = openSync(tempFile, "w");
+    try {
+      writeFileSync(handle, JSON.stringify(config, null, 2));
+      fsyncSync(handle);
+    } finally {
+      closeSync(handle);
+    }
+    renameSync(tempFile, mainFile);
```
zbyju: `rv-logic` 🟡 **The rename itself is not fsynced.**

`fsyncSync(handle)` makes the temp file's bytes durable. The directory entry from `renameSync` is not synced, and ext4 commits directory changes on its own schedule, up to 5 s later. A power cut soon after the 200 replays the journal without the rename: the box boots on the old config and `main.json.tmp` holds the new one. A lost update, not corruption.

Opening a directory for fsync fails on Windows, and `base.json` ships `pc.os: "windows"`, so guard it.

```suggestion
    renameSync(tempFile, mainFile);
    if (process.platform !== "win32") {
      const dir = openSync(configDir, "r");
      try {
        fsyncSync(dir);
      } finally {
        closeSync(dir);
      }
    }
```

source/apps/configer/src/services/db/main.ts
```diff
@@ -1,40 +1,135 @@
...
+ */
+function loadStored(
+  mainFile: string,
+  backupFile: string,
+  corruptFile: string
+): unknown {
+  if (!existsSync(mainFile)) return undefined;
+
+  const stored = parseFile(mainFile);
+  if (stored !== undefined) return stored;
```
zbyju: `rv-logic` 🟡 **Boot accepts any JSON, not only an object.**

`loadStored` returns anything that parses. `update()` rejects a string or an array at line 109 because `lodash.merge` spreads them over the defaults; boot and the backup guard at line 84 have no such check.

1. `main.json` holds `"abc"` (valid JSON, a string).
2. `parseFile` returns it, `merge(freshBase(), "abc")` adds keys `0`, `1`, `2`.
3. `write()` sees the file parses, so it copies it over `main.json.bak`.
4. The box serves a config with index keys and the good backup is gone.

Fix: reject a non-plain-object here and in the line 84 guard, the same way `update()` does.

source/apps/configer/src/services/db/main.ts
```diff
@@ -1,40 +1,135 @@
...
+      closeSync(handle);
+    }
+    renameSync(tempFile, mainFile);
+  }
+
+  // Boot is not validated on purpose; an odd stored value must not stop the box.
+  let data = merge(
+    freshBase(),
+    loadStored(mainFile, backupFile, corruptFile)
+  ) as MainConfig;
```
zbyju: `rv-logic` · `rv-types` · `rv-tests` 🟡 **Boot accepts what PUT rejects, and nothing locks that in.**

The stored file is cast to `MainConfig` with no check, and the comment says that is on purpose. Agreed: a bad stored value must not stop the box. But the UI this PR prepares for will GET this config and PUT it back.

1. A box's `main.json` holds `cameraType: "Dahua IPC"`. The panel matches with `includes()`, so it works today.
2. Boot is unvalidated, the box runs fine.
3. The UI GETs the config, the user edits only `babybox.name`, PUTs the whole thing back.
4. 400 on `camera.cameraType`. Nothing saves until someone edits the file by hand.

Also, every current boot test supplies well-typed data or unparseable JSON. A later "validate at boot and throw" change would pass the suite.

Fix: call `validateMainConfig(data)` after this merge and `console.warn` each error, keep booting either way. Add one boot test with a wrong-typed stored field (for example `port: "8080"`) that asserts `mainConfig()` still resolves.

source/apps/configer/src/types/main.types.ts
```diff
@@ -9,175 +33,172 @@ export interface MainConfig {
...
+    checkString(errors, camera.ip, "camera.ip");
+    checkString(errors, camera.username, "camera.username");
+    checkString(errors, camera.password, "camera.password");
+    checkInteger(errors, camera.updateDelay, "camera.updateDelay");
+    checkOneOf(
+      errors,
+      lowerCased(camera.cameraType),
+      "camera.cameraType",
+      cameraTypes
+    );
```
zbyju: `rv-types` · `rv-sweep` 🟡 **cameraType: the type, the validator and the panel disagree.**

- The type `MainConfigCameraType` (line 78) allows lower case only.
- This check accepts any case, so `"DAHUA"` passes and is stored typed as `"dahua"`. A select filled from `cameraTypes` shows no selection for that box.
- The panel (`source/apps/panel/src/utils/panel/camera.ts`) accepts any string that contains a known name and falls back to dahua, so `"DAHUA 2MP"` works in the panel today but this check rejects it with 400.

`docs/decisions.md` justifies the case-insensitive compare with "a stricter check would reject a config that works today". The check is still stricter than the panel in the other direction.

Fix: normalise `cameraType` to a known name (lower-case, substring match) on write and on load, so the stored value matches the type and every value the panel accepts passes. Or widen the field to `string` and export a `normalizeCameraType()` helper.

source/apps/configer/src/types/main.types.ts
```diff
@@ -9,175 +33,172 @@ export interface MainConfig {
...
+  if (typeof value !== "string") errors.push({ path, msg: "must be a string" });
+}
+
+function checkInteger(
+  errors: ConfigError[],
+  value: unknown,
+  path: string
+): void {
+  if (!Number.isInteger(value)) errors.push({ path, msg: "must be an integer" });
+}
```
zbyju: `rv-logic` 🟡 **checkInteger has no range.**

Any integer passes, including 0, negatives and ports above 65535. This is the check meant to stop a bad config reaching disk.

1. A UI number field sends `configer.port` as 70000.
2. `Number.isInteger` passes, the config is written.
3. The box restarts, `app.listen(70000)` throws ERR_SOCKET_BAD_PORT.
4. Configer never starts, someone must go on site.

A negative `units.requestDelay` also gives the panel a poll loop with no delay, since `requestDelay || 2000` only catches 0.

Fix: add `min`/`max` parameters, use 1..65535 for the two ports and a positive minimum for delays, timeouts and thresholds.

`rv-types` 🟢 The interface fields are typed `number` but rejected when fractional. Note the integer rule on the fields, or accept a finite number where a fraction makes sense (`units.voltage.*`).

source/apps/configer/src/types/main.types.ts
```diff
@@ -9,175 +33,172 @@ export interface MainConfig {
...
+
+function lowerCased(value: unknown): unknown {
+  return typeof value === "string" ? value.toLowerCase() : value;
+}
+
+/*
+ * The one shape check for main.json. Returns every problem it finds, each with a
+ * dotted path, so a caller can point at the field that is wrong.
+ */
+export function validateMainConfig(config: unknown): ConfigError[] {
```
zbyju: `rv-types` 🟡 **Unknown keys pass validation and are stored.**

The check reads known paths and ignores the rest, so what it accepts is wider than `MainConfig`. `PUT /config/main` has no auth and `main.json` sits on box flash.

1. PUT `{"babybox":{"name":"x"},"junk":"<1 MB string>"}`.
2. No error; `merged` is written to `main.json` and kept in memory.
3. Every later boot merges and rewrites the junk.

Fix: build the result from the known keys and return it, or push an error for each root key not in the `MainConfig` key list.

source/apps/configer/src/services/db/main.ts
```diff
@@ -1,40 +1,135 @@
...
+      return { ok: false, errors: [{ path: "", msg: "must not be empty" }] };
+    }
 
-  // Merge base to db
-  const merged = merge(base, db.data);
-  db.data = merged;
-  await db.write();
+    const merged = merge(freshBase(), body) as MainConfig;
+    const errors = validateMainConfig(merged);
+    if (errors.length > 0) return { ok: false, errors };
```
zbyju: `rv-brief` 🟡 **PUT fills omitted fields from base.json, not the current config.**

This is the documented and tested design, and the comment above marks PUT as a full replace. Nothing in the endpoint enforces "full body" though: any non-empty object is accepted.

1. A box has `units.engine.ip = 10.1.1.99` on disk.
2. A client PUTs only `{ "babybox": { "name": "Brno" } }`.
3. The body is merged over `base.json`, not over the stored config.
4. `main.json` is written with `units.engine.ip` reset to the default.

A settings form naturally produces this shape of body. Zero live impact today. Fix: state in the route that PUT needs the full config, or reject a body missing top-level keys until PATCH (P2) exists.

`rv-types` 🟡 **validateMainConfig is not a type guard.** `merged` is already typed `MainConfig` on line 121, one line before anything is checked. A second write path can cast the same way and never call the validator; the compiler stays silent. Fix: `parseMainConfig(value: unknown): { ok: true; config: MainConfig } | { ok: false; errors: ConfigError[] }` and drop the `as MainConfig`.

source/apps/configer/src/services/db/main.ts
```diff
@@ -1,40 +1,135 @@
...
+  dirname(fileURLToPath(import.meta.url)),
+  "../../../configs"
+);
+
+function parseFile(file: string): unknown {
+  try {
+    return JSON.parse(readFileSync(file, "utf-8")) as unknown;
+  } catch {
+    return undefined;
+  }
```
zbyju: `rv-logic` 🟡 **parseFile hides read errors as parse errors.**

The catch covers `readFileSync` too, not only `JSON.parse`.

1. Someone edits `main.json` as root; it ends up mode 600.
2. Configer runs as another user, `readFileSync` throws EACCES.
3. `parseFile` returns undefined, the log says "does not parse".
4. `copyFileSync` at line 59 throws the same EACCES. `index.ts` awaits `getMainDb()` before `listen`, so configer does not start.

Fix: read and parse in separate steps. Treat only a `JSON.parse` failure as corrupt. On a read failure, log the errno and fall back without writing `main.json.corrupt`.

source/pnpm-lock.yaml
```diff
@@ -1827,7 +1829,7 @@ packages:
   /call-bind/1.0.2:
     resolution: {integrity: sha512-7O+FbCihrB5WGbFYesctwmTKae6rOiIzmz1icreWJ+0aA7LJfuqhEso2T9ncpcFtzMQtzXf2QGGueWJGTYsqrA==}
     dependencies:
-      function-bind: 1.1.1
+      function-bind: 1.1.2
```
zbyju: `rv-sweep` 🟡 **Lockfile changed beyond adding vitest.**

Adding vitest needs only the new `/vitest/0.9.4` and `/vite/2.9.14` entries. This diff also drops `/function-bind/1.1.1`, repoints `call-bind`, `get-intrinsic` and `has` to 1.1.2, removes `dev: true` from `/function-bind/1.1.2`, and adds a `deprecated:` note on `/loupe/2.3.4`. That is pnpm re-resolving `^1.1.1` against today's registry, which `docs/learnings.md` warns against, and `docs/decisions.md` says "the lockfile gains no new package versions".

Boxes run `pnpm run build`, which starts with a non-frozen `pnpm install`. `startup/src/logic/start/ubuntu.js:83` treats any stderr from that step as a failed update and skips `override()`. Whether pnpm 7.5.0 prints the new deprecation warning to stderr or stdout is the one thing to verify.

Fix: restore `/function-bind/1.1.1` and its three references, keep only the vitest and vite additions.

source/apps/configer/src/services/db/main.ts
```diff
@@ -1,40 +1,135 @@
...
+    /*
+     * Only back up a file we could read. A corrupt main.json must never
+     * overwrite a good main.json.bak.
+     */
+    if (parseFile(mainFile) !== undefined) copyFileSync(mainFile, backupFile);
+
+    const handle = openSync(tempFile, "w");
+    try {
+      writeFileSync(handle, JSON.stringify(config, null, 2));
+      fsyncSync(handle);
```
zbyju: `rv-tests` 🟡 **fsync-before-rename has no test.**

Delete this line and every test still passes. "leaves no temp file behind" cannot tell. Durability across a power cut is not observable from a unit test, so the only test that catches a dropped `fsyncSync` spies on it.

Fix: spy on `fsyncSync` and `renameSync` from `node:fs`, call `update()`, assert `fsyncSync` was called before `renameSync`.

source/apps/configer/src/routes/configRoute.ts
```diff
@@ -15,13 +14,12 @@ router.get(["/version", "/versions"], async (req: Request, res: Response) => {
...
     return res
       .status(400)
-      .json({ msg: `${JSON.stringify(c)} is not a type of 'MainConfig'` });
+      .json({ msg: "Body is not a valid MainConfig", errors: result.errors });
   }
-  const main = await DbFactory.getMainDb();
-  await main.update(c);
-  return res.json(main.data());
+  return res.json(result.config);
 });
```
zbyju: `rv-tests` 🟡 **PUT /main has zero tests.**

No test file exists for this route. Neither the 400 `{ msg, errors }` shape nor the 200 body is checked. The success path also changed in this PR, from echoing `main.data()` to `result.config`.

Fix: add `configRoute.test.ts`: PUT an invalid body, expect 400 with `{ msg, errors }`; PUT a valid body, expect 200 with the updated config.

source/apps/configer/src/services/db/main.ts
```diff
@@ -1,40 +1,135 @@
...
+    const errors = validateMainConfig(merged);
+    if (errors.length > 0) return { ok: false, errors };
 
-  async function update(c2: MainConfig): Promise<MainConfig> {
-    db.data = c2;
-    await db.write();
-    return db.data;
+    // Disk first: memory must never hold a config the disk does not have.
+    write(merged);
+    data = merged;
```
zbyju: `rv-tests` 🟡 **Write-before-memory ordering has no test.**

`docs/decisions.md` records "disk first" as a deliberate P0 fix. No test proves `data()` keeps the old config when `write()` throws, so the two lines can be swapped back without a failure.

Fix: mock one write-path fs call to throw, assert `update()` rejects and `data()` is unchanged.

docs/mottos.md
```diff
@@ -0,0 +1,22 @@
+# Mottos
```
zbyju: `rv-brief` 🟢 `mottos.md`, `learnings.md` and `config-ui-agent-workflow.md` (151 lines) are not in the PR body's "What changed" list. Only `decisions.md` is referenced. Add one bullet for the three.

zbyju:
## Review · 7 agents

### WHAT
`PUT /config/main` now validates every field (unit IPs and `startup` were unchecked before), writes `main.json` through a temp file, `fsync` and rename, keeps one `main.json.bak`, and returns field-level errors. A missing, empty, or corrupt config no longer stops the box from starting.

### RISK — medium
- **Blast radius** — `write()` runs on every configer boot on every deployed box.
- **Data** — a partial PUT merges over `base.json`, not the current config, so omitted fields reset to defaults. Zero live impact today, nothing calls the endpoint.

### WHY
P0 of `docs/plans/config-ui.md`. A config UI will be built on this endpoint. Today a bad PUT or a power cut mid-write can leave a box without a readable config, which needs someone on site.

### HOW
Route → `DbFactory.getMainDb()` → `update()` → `validateMainConfig` → `write()`.
```
PUT /main (configRoute.ts)
  -> main.update(req.body)
      reject if not a plain object, or empty
      merge(freshBase(), body)                    NOT merged with current data
      validateMainConfig(merged) -> on error { ok:false, errors }, nothing written
      write(merged): copy main.json -> .bak (only if it parses), write .tmp, fsync, rename
      data = merged -> { ok:true, config }
Boot (mainConfig(configDir))
  loadStored(): parse main.json; on failure copy to .corrupt, warn, try .bak
  merge(freshBase(), stored)                     unvalidated on purpose
  write(data)                                    runs on every boot
```

### TOUCHES
**Tests** — 28 added, 0 changed
- ✓ Boot writes `main.json` from `base.json` when none exists, keeps stored values, falls back to `.bak` on a corrupt file, keeps `.bak` untouched, keeps the corrupt file as `.corrupt`, falls back to `base.json` when both are corrupt.
- ✓ Update fills a partial body from `base.json`, rejects invalid, non-object and empty bodies without writing, keeps the previous content in `.bak`, leaves no temp file.
- ✓ Validator accepts `base.json`, rejects missing `engine`/`thermal`, non-string unit IP, missing `startup`, non-integer `refreshRequestLimit`, unknown camera type, wrong-case or unknown `pc.os`, non-object root. Accepts all 5 camera types and upper case.
- ✗ P0 "write atomically with fsync" has no test that `fsyncSync` runs before the rename.
- ✗ P0 "field-level errors from PUT" has no test of the Express route's 400 or 200 body.

**Schema** — none. **Events** — none.

**Praise** (`rv-sweep`) — `main.ts:84` backs up only a `main.json` that parses. One line, and it removes the whole class of bug where recovering from a corrupt file overwrites the last good copy.

---

Findings are inline on the diff: 0 🔴 · 15 🟡 · 2 🟢. The `rv-doubt` comment on `main.ts:101` is a design question that needs a decision, not a patch.

Agents: `rv-brief` · `rv-hygiene` (clean) · `rv-doubt` · `rv-logic` · `rv-types` · `rv-tests` · `rv-sweep`. Skipped `rv-structure` · `rv-events` · `rv-react` · `rv-styles` (no `.claude/review.json`). The 10 threads from the previous round were checked: all fixes are on the branch, none re-raised.

source/apps/configer/src/routes/configRoute.ts
zbyju: **Addressed in** [`081e289`](https://github.com/zbyju/babybox/commit/081e289)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- update() returns write-failed, nothing throws out of it
- route answers 500 with the error message
- memory and disk keep the old config

source/apps/configer/src/services/db/main.ts
zbyju: **Addressed in** [`caaa33f`](https://github.com/zbyju/babybox/commit/caaa33f)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- directory fsynced after the rename, skipped on win32
- your suggestion, pulled into a syncDir() helper

source/apps/configer/src/services/db/main.ts
zbyju: **Addressed in** [`0b9e76e`](https://github.com/zbyju/babybox/commit/0b9e76e)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- stored file must be a plain object, else it is corrupt
- same rule guards the .bak copy in write()
- test: main.json holding a string boots from .bak, .bak kept

source/apps/configer/src/services/db/main.ts
zbyju: **Addressed in** [`8799be1`](https://github.com/zbyju/babybox/commit/8799be1)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- boot warns once per field validateMainConfig rejects
- still boots; test with port "8080" asserts it resolves

source/apps/configer/src/types/main.types.ts
zbyju: **Addressed in** [`1e2d2f4`](https://github.com/zbyju/babybox/commit/1e2d2f4)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- check uses the panel's substring rule, "Dahua IPC" passes
- decisions.md entry updated
- rewriting the stored value or widening the type waits for a decision

docs/decisions.md records the type gap as deliberate and refuses to rewrite a typed value. Closing it flips that decision, so the thread stays open until it is made.

source/apps/configer/src/types/main.types.ts
zbyju: **Addressed in** [`490f8c8`](https://github.com/zbyju/babybox/commit/490f8c8)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- ports 1..65535; delays, timeouts, thresholds, refreshRequestLimit at least 1
- voltage divider and multiplier at least 1, addition any integer
- integer rule noted on MainConfig; fractions kept out, base.json has none

source/apps/configer/src/types/main.types.ts
zbyju: **Addressed in** [`2026b80`](https://github.com/zbyju/babybox/commit/2026b80)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- unknown keys reported at root and in every known section
- startup stays open, it has no known shape
- tests for a root key, a nested key, and startup

source/apps/configer/src/services/db/main.ts
zbyju: **Addressed in** [`ab4586e`](https://github.com/zbyju/babybox/commit/ab4586e)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- route comment: PUT is a full replace, missing keys get base.json defaults
- parseMainConfig returns the typed config or the errors
- update() uses it, no cast left on the write path

Rejecting a body that misses top-level keys would flip the recorded merge-over-base decision and its test, so the route documents the rule instead.

source/apps/configer/src/services/db/main.ts
zbyju: **Addressed in** [`0b9e76e`](https://github.com/zbyju/babybox/commit/0b9e76e)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- read and parse are separate steps
- read error logs the message, no .corrupt copy, falls back
- test mocks readFileSync to throw EACCES

source/pnpm-lock.yaml
zbyju: **Addressed in** [`bd609e9`](https://github.com/zbyju/babybox/commit/bd609e9)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- function-bind 1.1.1 and its three references restored
- dev: true back on 1.1.2, loupe deprecation note dropped
- frozen install with pnpm 7.5.0 passes
- diff now holds only the importer entry, vite 2.9.14, vitest 0.9.4

source/apps/configer/src/services/db/main.ts
zbyju: **Addressed in** [`caaa33f`](https://github.com/zbyju/babybox/commit/caaa33f)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- node:fs mocked with pass-through spies in main.test.ts
- test asserts fsync, then rename, then directory fsync

source/apps/configer/src/routes/configRoute.ts
zbyju: **Addressed in** [`0c6e143`](https://github.com/zbyju/babybox/commit/0c6e143)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- configRoute.test.ts: 400 with field errors, 200 then GET, 500 on write failure
- DbFactory mocked to a temp config dir, real express app on port 0

source/apps/configer/src/services/db/main.ts
zbyju: **Addressed in** [`081e289`](https://github.com/zbyju/babybox/commit/081e289)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- test makes fsyncSync throw ENOSPC through a node:fs mock
- asserts write-failed, data() and main.json unchanged

update() no longer rejects on a failed write. It returns write-failed, so the test asserts that instead.

docs/mottos.md
zbyju: **Addressed in** — the PR description, no commit
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- Docs section added, one bullet per file
- stale lines on the result shape, 400/500 and the test count corrected

source/apps/configer/src/services/db/main.ts
zbyju: **Addressed in** [`8ca7791`](https://github.com/zbyju/babybox/commit/8ca7791f2a7224de2a79e26b3fbc7e1720de0de4)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- boot merges in memory only, never writes
- `.bak` is now the config before the last PUT
- `main.json.corrupt` and its gitignore line removed
- guard kept: a corrupt `main.json` still never overwrites `.bak`
- test added: PUT, reboot, `.bak` unchanged
- new entry in `docs/decisions.md`, `.corrupt` entry marked superseded

Confirmed: nothing outside this repo reads or writes `main.json`. The guard stays because a box that booted from `.bak` can still get a PUT while `main.json` is corrupt.

source/apps/configer/src/types/main.types.ts
zbyju: **Addressed in** [`e460644`](https://github.com/zbyju/babybox/commit/e46064406826cd3809dd0620a152d1661249c633)
**Status:** done
**Posted by:** AI (not a human). Message me in Slack in case you need a human reply.

- check is an exact match against the lower-case list
- type, check and stored value now agree
- `DAHUA` and `Dahua IPC` get a 400
- `docs/decisions.md` entry rewritten

Decision: every deployed box holds one of the exact lower-case names, so nothing has to rewrite a typed value. The panel stays wider than the check on purpose.

PR #86, opened 2026-09-12
feat: share one config schema across the apps

zbyju:
Phase **P1** of [docs/plans/config-ui.md](docs/plans/config-ui.md): one zod schema for
the config shape, shared by configer, the panel and the backend.

## Why

The shape of `main.json` was written out five times: configer's types and its
hand-written validator, the panel's types, its config guard and its defaults. They had
already drifted apart. `camera.cameraType` was a union in one place and a plain
`string` in two. The panel's default voltage divider was `3400` where `base.json` says
`63`. The panel carried an `app.version` that nothing ever set. Every new field meant
four edits, and the config UI adds a fifth consumer of the shape.

The decision and the constraints behind it are in
[docs/decisions.md](docs/decisions.md), entry "The config shape is one zod schema in a
shared package".

## What changed

**New package** `source/packages/config-schema` (`@babybox/config-schema`)

- ESM, built by `tsc` to `dist/`, `main` and `types` point there. `zod` pinned to
  exactly `3.23.8` — the last 3.x before zod shipped v4 alongside v3, and the last one
  that runs on TypeScript 4.7, which all three apps use.
- `mainConfigSchema` keeps every rule the hand-written validator had: the same integer
  ranges, the same exact `cameraType` and `pc.os` lists, unknown keys rejected in every
  object, `startup` any object, `app.refreshRequestLimit` optional.
- `MainConfig` is `z.infer` of the schema. `validateMainConfig` and `parseMainConfig`
  wrap `safeParse` and keep the `{ path, msg }` error shape with a dotted path. zod
  reports the unknown keys of one object as a single issue, so they are expanded into
  one error per key.
- `defaultConfig()` holds the values `base.json` holds, and returns a fresh object each
  call so no caller can change what the next one reads.
- `"types": []` in its tsconfig: the package uses no Node and no test API, so its build
  cannot be broken by an unrelated `@types` package somewhere above the checkout.
- 37 tests, one per rejected field class plus the accepted shapes.

**Configer**

- `src/types/main.types.ts` is now a thin re-export of the package. The ~290 lines of
  hand-written checks are gone.
- `db/main.ts` and the route behave exactly as before. `PUT /config/main` still answers
  `400 { msg, errors: [{ path, msg }] }`. `isPlainObject` moved into `db/main.ts`, its
  only caller.
- The per-field tests moved to the package. What is left in configer is what only
  configer can check: `configs/base.json` deep-equals `defaultConfig()`, and it is a
  valid config.

**Panel**

- `types/panel/config.types.ts` derives every section from `MainConfig`. `Config` is
  the five sections the panel reads.
- `utils/panel/instanceCheck.ts` — the config guard is the shared schema now; the
  versions guard is untouched.
- `defaults/config.default.ts` returns the schema's defaults. This fixes the voltage
  divider (`3400` -> `63`) and the camera type (`"DAHUA"` -> `"dahua"`).
- `app.version` is dropped. Nothing set it and nothing read it — the only two mentions
  in the whole app were its own declaration and the default `"unknown"`.
- `utils/panel/camera.ts` — the `CameraType` enum and `stringToCameraType` are gone.
  The url lookup is a `Record` over the schema's union, so the compiler makes it
  exhaustive. `avm` keeps the avtech url. The second, unused `CameraType` enum in
  `types/panel/main.types.ts` is deleted with them.
- `panelLoop.getConfig()` returns `unknown`, which is what an HTTP body is; the guard
  is the only thing that turns it into a config.

**Backend**

- `src/types/config.types.ts` is `export type { MainConfig } from "@babybox/config-schema"`.
  No entry in the backend's `package.json`: startup copies `apps/backend/dist` and the
  backend's `package.json` to the repo-root `dist/` and runs `pnpm install` there,
  outside the workspace, where `workspace:*` cannot resolve. The type comes through a
  tsconfig `paths` entry that points at the package's built `.d.ts`.
- Checked: `grep -r config-schema apps/backend/dist` finds nothing, and
  `dist/types/config.types.js` is an empty module.

**Wiring**

- Root `build` and `dev` build the package first. CI gets a "Build config schema" step
  before the app builds and a "Config schema tests" step.
- Lockfile written with `pnpm@7.5.0` on Node 18. Still `lockfileVersion: 5.4`, and the
  diff is only the new importer, `zod@3.23.8` and the two `link:` entries.

## Decisions made

Written up in [docs/decisions.md](docs/decisions.md):

- **The panel accepts a config with a key it does not know.** The panel's guard is the
  shared schema with one rule relaxed: a value whose only problems are unrecognised
  keys still passes. If the panel refuses a config it never mounts, so an extra key in
  a deployed `main.json` would leave a box showing nothing until someone drives out to
  it. `PUT` still refuses that same key — rejecting belongs on the write, not the read.
  What this does **not** cover: any other wrong value (a bad type, a camera type
  outside the list) still stops the panel where the old loose guard let it through.
  decisions.md records that every deployed box was confirmed to hold an exact
  lower-case camera type on 2026-09-12.
- **`defaultConfig` is a function, not a constant,** so a pinia store cannot mutate the
  defaults other callers read.
- **The schema package declares no ambient types** (`"types": []`).

## Deliberately left out

- **Form metadata** (Czech labels, widget, unit suffix, hint, apply tier) — **P3**,
  where it is first read. Nothing in P1 or P2 uses it (motto 1). The checklist item was
  already moved there in the plan.
- **`PATCH /config/main`** for a partial update — **P2**.
- **Auth on the write endpoint** — **P5**. Still unauthenticated, still unused.
- The configer write path, the route and the on-disk `main.json` format are unchanged.

## How to test

```bash
cd source
npx -y -p node@18.12.1 -p pnpm@7.5.0 sh -c 'pnpm install --frozen-lockfile && pnpm run build'
npx -y -p node@18.12.1 -p pnpm@7.5.0 pnpm -F @babybox/config-schema exec vitest run
npx -y -p node@18.12.1 -p pnpm@7.5.0 pnpm -F babybox-panel-configer exec vitest run
npx -y -p node@18.12.1 -p pnpm@7.5.0 pnpm -F babybox-panel exec vitest run --environment jsdom
npx -y -p node@18.12.1 -p pnpm@7.5.0 pnpm -F babybox-panel-backend exec jest
```

Results on Node 18.12.1: schema 37 passed, configer 24 passed, panel 26 passed,
backend 118 passed.

By hand: start configer, `PUT` a body with `units.engine.ip` set to a number, and check
the 400 still lists `units.engine.ip` with `must be a string`. `PUT` a body with an
unknown key under `camera` and check each unknown key gets its own error. Then start
the panel against that configer and check it still mounts.

## Upgrade safety, checked

- **Clean clone, Node 18.12.1:** `git clone` of this branch into an empty directory,
  then `pnpm install --frozen-lockfile && pnpm run build` from `source/`. Exit 0,
  **stderr empty** — which is what the startup app requires; any stderr from
  `pnpm run build` fails the update on the box.
- **Lockfile:** `lockfileVersion: 5.4`, written by `pnpm@7.5.0` under Node 18.
  `--frozen-lockfile` passes. The diff adds the `packages/config-schema` importer,
  `zod@3.23.8` and two `link:` lines, nothing else. `vitest` is `^0.9.3`, the version
  configer already had, so no new version is downloaded.
- **Build order:** the package is built before every app that reads it, in the root
  `build` script (what a box runs after `git pull`), in `dev` and in CI.
- **`main.json` unchanged.** The schema declares its keys in `base.json`'s order, so a
  config that goes through a `PUT` is written back in the same order it had.
- **The backend's standalone install is untouched:** no new dependency in its
  `package.json`, nothing required from the package in its `dist`.
- **Panel bundle:** 228.04 KiB -> 285.27 KiB raw, 78.63 KiB -> 92.51 KiB gzipped. That
  is the zod runtime, and it was the known price of this decision.
- **Panel typecheck:** `vue-tsc -p tsconfig.vitest.json` reports the same 20 errors as
  `origin/main`, measured both ways on this branch. It is red on main already; this
  change adds nothing to it.
- **Lint** runs clean the way CI runs it, for both the backend and the panel.

## One fix that is not P1

CI is red on `main` today: three `configRoute` tests fail with `TypeError: fetch failed`
(run [34687208149](https://github.com/zbyju/babybox/actions/runs/34687208149), the merge
of #85). The cause is in the test, not in configer: it sets a `connection: close`
request header, which the fetch spec forbids, and undici on Node 18.12.1 — the version
CI and every box run — fails the whole request with `invalid connection header`. Node 24
ignores it, which is why it was green when it was written.

The header is dropped. `afterEach` already calls `server.closeAllConnections()`, so it
bought nothing. Configer's suite passes on Node 18.12.1 now. Written up in
[docs/learnings.md](docs/learnings.md). Shout if you would rather have this in its own
PR.


source/apps/panel/src/utils/panel/instanceCheck.ts
```diff
@@ -1,30 +1,15 @@
...
-    isInstanceOfUnitsConfig(object.units)
-  );
-};
+/*
+ * The whole config file, checked against the shared schema. The panel only reads
+ * five sections, but configer sends all of them and a wrong value anywhere is a
+ * config the maintainer has to see.
+ */
+export const isInstanceOfConfig = (object: unknown): object is Config =>
+  isMainConfig(object);
```
zbyju: **blocker** · The panel's read guard went from "are the keys there" to every schema rule except unknown keys. Configer boot only warns on a bad stored value and serves it as-is (`services/db/main.ts` L148-151), and no write-path validator existed before #85, so deployed `main.json` files were never checked against these rules.

A value the old panel handled on purpose now stops the box. Checked against the built package, each of these makes `isMainConfig` return `false`:

- `app.refreshRequestLimit: 0` or negative. `checkRefreshLimit` in `panelLoop.ts` has an explicit "disable refresh if limit is 0 or negative" branch, shipped with the feature in c2ff044. That is the only way to turn the reload off, so a box with `0` is plausible.
- `units.requestDelay: 0`, `warningThreshold: 0`, `errorThreshold: 0`, `camera.updateDelay: 0`. The panel falls back with `|| 2000`, `|| 5`, `|| 25`, `|| DEFAULT_UPDATE_DELAY` (`panelLoop.ts` L296, `state.ts` L17-19, `useCamera.ts` L60).
- `refreshRequestLimit: null`, `requestDelay: 2000.5`, `backend.port: "5000"`, `pc.os: "linux"`. The panel never reads `pc.os` or the ports, yet the whole config is refused.

Failure path: `initializeConfig` throws "Config file error" on every retry, `AppState` never reaches `Ok`, `AppState.vue` renders the router only in that state, so `MainView` never mounts, `startPanelLoop` never runs and the engine watchdog stops. The box blocks itself until someone drives out (motto 3).

The decision "The panel accepts a config with a key it does not know" already says rejecting belongs on the write path, and lists this exact risk under "Gave up". Two ways out:

1. Make the panel's check structural: every section present and an object, primitives of the type the panel reads. `console.warn` the full `validateMainConfig` result so a maintainer still sees the bad value. `PUT /config/main` keeps the strict check.
2. Or run `validateMainConfig` on `GET /config/main` from every deployed box, confirm only unknown-key errors come back, and record it in decisions.md next to the camera-type entry.

Option 1 is cheaper and does not depend on a survey staying true after the next hand edit.

Related nit: if the panel's check stays identical to `isMainConfig`, this wrapper is `isMainConfig` under another name (`Config` is a `Pick` of `MainConfig`) and its four spec tests copy the package's one for one; import `isMainConfig` in `panelLoop.ts` and delete the wrapper. If the check differs as in option 1, this wrapper is where the difference lives, keep it.

zbyju (commented):
Review stage of [config-ui-agent-workflow.md](docs/plans/config-ui-agent-workflow.md), phase P1. Four lenses (correctness, upgrade safety, simplicity and types, tests), each finding then checked by a separate refuter against the code. 26 raised, 17 kept, 9 refuted.

Kept: 1 blocker, 4 should, 12 nit, as inline comments below.

Refuted, for the record: two errors on one path for a non-integer out of range (no written contract says one per field); thinner range coverage than a full table (matches the old suite case for case); `toEqual` vs `toStrictEqual` on the base.json test; unused-outside exports in the package index; the `files` field in package.json; the first build needing the registry (pre-existing, every update already resolves the backend's ranges online); four of seven comment nits.

Not an approve or a request for changes; humans do that.

source/packages/config-schema/src/validate.test.ts
```diff
@@ -0,0 +1,281 @@
...
+
+  it("reports every unknown key of an object on its own", () => {
+    const value = config();
+    value.pc = { ...(value.pc as Fields), first: 1, second: 2 };
+
+    expect(validateMainConfig(value)).toEqual([
+      { path: "pc.first", msg: "unknown key" },
+      { path: "pc.second", msg: "unknown key" },
+    ]);
+  });
```
zbyju: **should** · Unknown-key rejection is asserted for the root, `babybox` and `pc` only. Nothing checks it for `backend`, `configer`, `units`, `units.engine`, `units.thermal`, `units.voltage`, `camera` or `app`. The `isMainConfig` test that adds `camera.zoom` asserts `true`, so it passes with or without strictness.

Checked: with `.strict()` removed from `unit`, `voltage`, `camera` and `app` in `schema.ts`, all 37 tests stay green. `validateMainConfig` then returns `[]` for `camera.junk`, `app.junk`, `units.engine.junk` and `units.voltage.junk`, and `parseMainConfig` returns `ok: true` with those keys stripped. So a `PUT` with a typo like `camera.usernme` would answer 200 and silently drop the value instead of the 400 the P1 brief requires ("unknown keys rejected in every object"). Configer's route and db tests do not cover unknown keys either.

Suggestion: one table test over every object path (`""`, `babybox`, `backend`, `configer`, `units`, `units.engine`, `units.thermal`, `units.voltage`, `camera`, `pc`, `app`). Set `<path>.junk = 1` on a fresh `defaultConfig()` and expect exactly `[{ path: "<path>.junk", msg: "unknown key" }]`. The three existing unknown-key tests can fold into it.

source/packages/config-schema/src/validate.test.ts
```diff
@@ -0,0 +1,281 @@
...
+    });
+  });
+});
+
+describe("isMainConfig", () => {
+  it("accepts a valid config", () => {
+    expect(isMainConfig(config())).toBe(true);
+  });
+
+  it("accepts a config with a key we do not know", () => {
```
zbyju: **should** · `isMainConfig` passes a value whose issues are *all* `unrecognized_keys`. No test covers a config that has an unknown key *and* another problem. Changing `every` to `some` keeps all four `isMainConfig` tests and all four `isInstanceOfConfig` tests green (checked in a scratch copy), but then a stored `main.json` with `junk: 1` and `units.engine.ip: 10` would mount the panel on a wrong-typed value. decisions.md ("The panel accepts a config with a key it does not know") says every other rule must still stop the panel, so this is the case worth pinning. It stays relevant whatever rule the guard ends up with after the blocker thread.

```ts
it("rejects a config with an unknown key and a wrong type", () => {
  const value = withUnits({ engine: { ip: 10 } });
  expect(isMainConfig({ ...value, junk: 1 })).toBe(false);
});

it("rejects a config with an unknown key and an unknown camera type", () => {
  const value = config();
  value.camera = { ...(value.camera as Fields), cameraType: "axis" };
  expect(isMainConfig({ ...value, junk: 1 })).toBe(false);
});
```

Both fail on the `some` mutant and pass on the real code.

source/apps/panel/src/utils/panel/camera.ts
```diff
@@ -1,21 +1,15 @@
...
-export const getURLPostfix = (type: CameraType): string => {
-  if (type === CameraType.dahua) return "/cgi-bin/snapshot.cgi?Channel=0/";
-  if (type === CameraType.hikvision)
-    return "/ISAPI/Streaming/channels/101/picture?snapShotImageType=JPEG";
-  if (type === CameraType.avtech)
-    return "/cgi-bin/guest/Video.cgi?media=JPEG&channel=0/";
-  if (type === CameraType.vivotek) return "/cgi-bin/viewer/video.jpg/";
-  return "ERROR";
+/* avm is an AVTECH camera sold under the older name, so it answers the same url. */
+const urlPostfixes: Record<MainConfigCameraType, string> = {
```
zbyju: **should** · No test covers this mapping, and the type does not cover it in practice either: the panel `build` runs `vue-tsc --noEmit` on a solution tsconfig that lists zero files (learnings.md), and `pnpm typecheck` is red and not in CI. So if a camera type is added to `cameraTypes` without a URL here, or `avm` is later pointed at another string, nothing goes red; the panel puts `undefined` or the wrong path into the snapshot URL and a real camera stops showing frames.

The P1 brief names "`avm` keeps the avtech URL" as an acceptance item, and today it is checked only by reading the diff. The five URLs do match the old if-chain byte for byte.

Suggestion: `camera.spec.ts` next to this file:

```ts
import { cameraTypes } from "@babybox/config-schema";
import { describe, expect, it } from "vitest";
import { getURLPostfix } from "./camera";

describe("getURLPostfix", () => {
  it("returns a path for every camera type", () => {
    for (const type of cameraTypes) expect(getURLPostfix(type)).toMatch(/^\//);
  });

  it("avm uses the avtech url", () => {
    expect(getURLPostfix("avm")).toBe(getURLPostfix("avtech"));
  });
});
```

CI already runs the panel's vitest, so this runs on every PR.

.github/workflows/ci.yml
```diff
@@ -39,9 +39,16 @@ jobs:
           pnpm -F babybox-panel exec eslint . --ext .vue,.js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts --ignore-path .gitignore
           pnpm -F startup exec eslint src/
 
+      # Vsechny aplikace ctou tvar konfigurace odsud, musi se sestavit jako prvni
+      - name: Build config schema
+        run: pnpm -F @babybox/config-schema build
+
       - name: Build
         run: pnpm -F babybox-panel build && pnpm -F babybox-panel-backend build && pnpm -F babybox-panel-configer build
```
zbyju: **should** · The rule "the backend's `dist` must not `require("@babybox/config-schema")`" (decisions.md, constraint 2) is only checked by hand. The tsconfig `paths` entry makes a value import compile clean. Checked in a scratch copy: `import { defaultConfig } from "@babybox/config-schema"` in `index.ts` builds with exit 0 and emits `require("@babybox/config-schema")`. No jest test loads `index.ts`, so CI stays green. On a box, startup copies `apps/backend/dist` and the backend's `package.json` to the repo-root `dist/` and runs `pnpm install` there; the module is not in that `package.json` and no ancestor `node_modules` has it, so `pm2 start babybox` fails with module-not-found on every box. P4 (`POST /reload`) is the first place someone will be tempted to validate with the schema in the backend.

Suggested: one step right after Build.

```yaml
# The backend dist is installed standalone on the box; the schema package must stay type-only
- name: Backend dist has no runtime schema import
  run: '! grep -rl "config-schema" apps/backend/dist'
```

The type-only re-export emits an empty module (its comment is dropped by `tsc`), so this grep is clean today, and it checks the artifact that actually ships.

source/apps/configer/src/types/main.types.test.ts
```diff
@@ -2,185 +2,28 @@ import { readFileSync } from "node:fs";
...
-      path: "startup",
-      msg: "must be an object",
-    });
+/*
+ * base.json is the file configer merges a stored config over, and defaultConfig is
+ * the same values for everyone who has no file to read. They must not drift apart.
+ * The rules themselves are tested in @babybox/config-schema.
+ */
+describe("base.json", () => {
+  it("is the schema's defaults", () => {
```
zbyju: **nit** · The schema comment and learnings.md say the keys are declared in `base.json` order so the copy that `safeParse` returns, which `update()` writes to disk, keeps the file order. No test checks that: `toEqual` ignores key order, and so does every other assert on the written file. Moving `pc` above `camera` in `schema.ts` keeps all tests green, and each box then rewrites its `main.json` in the new order on its first PUT.

```ts
it("keeps the key order of base.json when parsed", () => {
  expect(JSON.stringify(parseMainConfig(baseConfig()).config)).toBe(
    JSON.stringify(baseConfig())
  );
});
```

Also, the learnings line reads as if the rewrite repeats; it happens once per box.

source/apps/configer/src/types/main.types.ts
```diff
@@ -1,288 +1,11 @@
...
-  if (errors.length > 0) return { ok: false, errors };
-  return { ok: true, config: value as MainConfig };
-}
-
 /*
- * The one shape check for main.json. Returns every problem it finds, each with a
- * dotted path, so a caller can point at the field that is wrong.
+ * The config shape lives in @babybox/config-schema, so configer, the panel and the
+ * backend describe it once. Configer reaches it through this file, so there is one
+ * import path for the shape inside the app.
```
zbyju: **nit** · This file is a pure re-export with one runtime importer, `services/db/main.ts`. The comment says it exists "so there is one import path for the shape inside the app", but with one importer there is nothing to unify, and `types/` now holds a file with no types. The backend's `types/config.types.ts` has the same shape for a real reason (type-only through a tsconfig `paths` entry); configer has a normal `workspace:*` dependency, so the extra hop buys nothing (motto 1).

The P1 brief did ask for "a thin re-export", so this pushes back on the brief. Suggestion: delete `main.types.ts` and its test file, import from `"@babybox/config-schema"` directly in `db/main.ts` (that import block is already in this diff), and move the two base.json tests into `services/db/main.test.ts` as a `describe("base.json")` block; that file already reads `repoBase`.

If you keep the file, rewrite the comment so it gives the actual reason.

source/apps/panel/src/defaults/config.default.ts
```diff
@@ -1,79 +1,27 @@
...
-  return {
-    name: "BABYBOX_NAME",
-  };
-};
+/*
+ * What the stores hold until configer answers. The values are the schema's defaults,
+ * which are the same ones configer falls back to, so the panel never shows a number
+ * no box would ever have.
+ */
+export const getDefaultBackendConfig = (): BackendConfig =>
```
zbyju: **nit** · After this rewrite every export here is one line that returns one section of `defaultConfig()`, and the only caller is `pinia/configStore.ts`, which casts each result to the type the function already returns (`getDefaultBackendConfig() as BackendConfig`). That is a file, five exports, an import block, a comment and five no-op casts for one call.

Suggest deleting this file and building the store state from one call. This typechecks with the panel's TS 4.7.4 and pinia 2.0.17 under strict mode, and `initialised` still infers as `boolean` without the `as boolean`:

```ts
import { defaultConfig } from "@babybox/config-schema";

state: () => {
  const { app, babybox, backend, camera, units } = defaultConfig();
  return { initialised: false, app, babybox, backend, camera, units };
},
```

If you do this, also update the "Panel defaults" row in `docs/plans/config-ui.md` (line 33), which still names this file.

docs/decisions.md
```diff
@@ -150,3 +150,74 @@ Context · Decision · Why · Gave up · Where
...
+  divider is 3400 where `base.json` says 63, and the panel has an `app.version` that
+  nothing sets. The [config-ui plan](plans/config-ui.md) left zod versus hand-written
+  guards open until P0 was merged.
+- Decision: one zod schema in `source/packages/config-schema`
+  (`@babybox/config-schema`). `MainConfig` is `z.infer` of it. Configer and the panel
+  validate with it, the backend imports the type only. zod pinned to `3.23.8`.
+- Why: one source for the type, the validation, the defaults and, in P3, the form
+  descriptor. `3.23.8` is the last 3.x before zod started shipping v4 next to v3
+  (3.25), and it runs on TypeScript 4.7, which all three apps use; zod 4 needs 5.5.
+- Constraints found on the way, each of them shapes the package:
```
zbyju: **nit** · Items 2 and 3 of "Constraints found on the way" repeat two learnings added in this PR almost word for word: "The backend's dist is installed standalone" (learnings.md, Startup) and "pnpm 7.5.0 cannot reach the registry on Node 20+" (learnings.md, Tooling). The pnpm one was already rewritten once in this PR, so a second copy will go stale. The file header asks for short entries.

Suggestion: keep item 1 (it follows from the decision). Shrink item 2 to the part that is a choice: "The backend imports the type only, through a tsconfig `paths` entry and `import type`; why in learnings.md, Startup." Drop item 3; the lockfile procedure lives in learnings.md, Tooling, and the P1 brief already links there.

docs/decisions.md
```diff
@@ -150,3 +150,74 @@ Context · Decision · Why · Gave up · Where
...
+
+- Context: the package exports the defaults that base.json holds on disk. The panel
+  puts them straight into a pinia store, which is reactive and mutable.
+- Decision: `defaultConfig()` builds a fresh object on every call.
+- Why: a shared constant that one store mutates changes what every later caller reads,
+  and nothing in the type system stops it. The cost is one object per call, at boot.
+- Gave up: nothing.
+- Where: `source/packages/config-schema/src/defaults.ts`.
+
+## 2026-09-12 — The schema package declares no ambient types
```
zbyju: **nit** · This entry, the Tooling learning "Backend `tsc` picks up `bun-types`" (learnings.md line 15) and the Shared package learning "A `tsc` build in a parent folder's shadow needs `"types": []`" (line 74) are three entries on one problem with two fixes. The entry says "Gave up: nothing", which is the sign of a fix, not a decision; the package imports only zod and its test imports vitest by name, so there is no trade-off to record.

Suggestion: delete this decision entry and merge the two learnings into one Tooling line: "`tsc` picks up `bun-types` from a parent `node_modules` (TS1005/TS1139 noise). A package that needs no ambient types sets `"types": []` in its tsconfig (config-schema does); the backend needs Node types, so pass `--typeRoots ./node_modules/@types` there."

docs/learnings.md
```diff
@@ -43,7 +47,37 @@ lesson: what happened, what to do instead.
...
+  carries a `keys` array, so expand it into one error per key before anyone sees it.
+- **`z.enum` needs an `errorMap` to carry a custom message.** `invalid_type_error`
+  only covers a value that is not a string; a string outside the list is
+  `invalid_enum_value` and keeps zod's own wording.
+- **zod's `safeParse` returns a copy, built in the order the schema declares.** The
+  written `main.json` follows the schema's key order, so declare the keys in the order
+  `base.json` has them or every box rewrites its file on the first PUT.
+- **A `tsc` build in a parent folder's shadow needs `"types": []`.** Cheaper and more
+  correct than `--typeRoots` on the command line when the package needs no ambient
+  types at all.
```
zbyju: **nit** · Configer and panel tests import `@babybox/config-schema` through the workspace link, and the package's `main` points at its gitignored `dist/`. There is no `pretest`, and only the root `build`/`dev` scripts and CI build the package. So on a fresh clone the configer and panel suites fail to resolve the package until it is built, and after an edit to `src/schema.ts` they run against the previous build and stay green.

Suggest one line under "Shared package": run `pnpm run build:schema` before the configer or panel tests, since they load the package from `dist/`, not `src/`.

docs/learnings.md
```diff
@@ -43,7 +47,37 @@ lesson: what happened, what to do instead.
...
+
+## Tests
+
+- **`fetch` on Node 18 refuses a request that sets `connection`.** undici calls it an
+  invalid connection header and the call fails with `TypeError: fetch failed`; Node 24
+  ignores the header, so a test written on Node 24 only goes red on CI. Let the agent
```
zbyju: **nit** · "Node 24 ignores the header" is not what happens. Checked on Node 24.20.0: the request goes through and the server receives `connection: close`. On Node 18.12.1 undici rejects it with `invalid connection header`. Suggest "Node 24 forwards the header" so the file stays a record of what was observed. Same wording in the PR body.

source/packages/config-schema/src/validate.ts
```diff
@@ -0,0 +1,61 @@
...
+/*
+ * Every problem with a value, each with a dotted path, so a caller can point at the
+ * field that is wrong. An empty list means the value is a MainConfig.
+ */
+export function validateMainConfig(config: unknown): ConfigError[] {
+  const result = mainConfigSchema.safeParse(config);
+  return result.success ? [] : toConfigErrors(result.error);
+}
+
+// The one place an unknown value becomes a MainConfig.
```
zbyju: **nit** · Two comments here can go.

L41 "The one place an unknown value becomes a MainConfig" is no longer true: `isMainConfig` (L53) is a type guard and also turns `unknown` into `MainConfig`, and the panel relies on it.

L32-35 above `validateMainConfig` repeats the signature and the `ConfigError` type. The only extra point (a path so a caller can point at a field) is already in the `toConfigErrors` comment above it.

source/packages/config-schema/src/validate.ts
```diff
@@ -0,0 +1,61 @@
...
+      msg: "unknown key",
+    }));
+  });
+}
+
+/*
+ * Every problem with a value, each with a dotted path, so a caller can point at the
+ * field that is wrong. An empty list means the value is a MainConfig.
+ */
+export function validateMainConfig(config: unknown): ConfigError[] {
```
zbyju: **nit** · `validateMainConfig` and `parseMainConfig` both run `safeParse` and `toConfigErrors`. Configer uses both (boot warning and PUT), and they agree on the errors only because each one repeats the same two steps. Make one call the other, so the boot warning and the PUT reply cannot drift:

```ts
export function validateMainConfig(config: unknown): ConfigError[] {
  const result = parseMainConfig(config);
  return result.ok ? [] : result.errors;
}
```

The other direction does not work, because `parseMainConfig` needs `result.data`.

source/apps/panel/src/utils/panel/camera.ts
```diff
@@ -1,21 +1,15 @@
...
 
-export const getURLPostfix = (type: CameraType): string => {
-  if (type === CameraType.dahua) return "/cgi-bin/snapshot.cgi?Channel=0/";
-  if (type === CameraType.hikvision)
-    return "/ISAPI/Streaming/channels/101/picture?snapShotImageType=JPEG";
-  if (type === CameraType.avtech)
-    return "/cgi-bin/guest/Video.cgi?media=JPEG&channel=0/";
-  if (type === CameraType.vivotek) return "/cgi-bin/viewer/video.jpg/";
-  return "ERROR";
+/* avm is an AVTECH camera sold under the older name, so it answers the same url. */
```
zbyju: **nit** · "avm is an AVTECH camera sold under the older name" is not something the repo or the brief establishes. What the old code showed (`includes("avtech") || includes("avm")` on main) is only that `avm` used the avtech URL. A wrong brand fact in a comment misleads the next person who debugs a camera.

Suggest: `/* avm cameras answer the avtech snapshot url. */`

source/apps/backend/src/types/config.types.ts
```diff
@@ -1,65 +1,6 @@
...
-
-export interface MainConfigPc {
-  os: "windows" | "ubuntu";
-}
-
-export interface MainConfigApp {
-  password: string;
-  refreshRequestLimit?: number;
-}
+/*
```
zbyju: **nit** · This comment and the one next to `paths` in `tsconfig.json` (L11-16) say the same thing: the backend's dist is installed outside the workspace, so there can be no dependency. Say it once. The tsconfig is the better place, because the odd part is the `paths` entry that points into `dist/index.d.ts`; a bare `export type` needs no explanation.

Keep the file (it follows `src/types/*.types.ts` like the other four), shorten this to one line:

```ts
// Type only, so nothing lands in the JS. Why there is no dependency: see paths in tsconfig.json.
export type { MainConfig } from "@babybox/config-schema";
```

source/apps/panel/src/types/panel/config.types.ts
```diff
@@ -1,57 +1,19 @@
...
+  MainConfig,
+  "app" | "babybox" | "backend" | "camera" | "units"
+>;
+
+export type AppConfig = MainConfig["app"];
+export type BabyboxConfig = MainConfig["babybox"];
+export type BackendConfig = MainConfig["backend"];
+export type CameraConfig = MainConfig["camera"];
+export type UnitsConfig = MainConfig["units"];
+export type UnitConfig = UnitsConfig["engine"];
```
zbyju: **nit** · `UnitConfig` has no importer left. The two functions that used it, `getDefaultUnitConfig` and `isInstanceOfUnitConfig`, are gone in this PR. Delete the line. (`VoltageConfig` below it is still used, keep that one.)

source/apps/panel/src/utils/panel/instanceCheck.ts
zbyju: Took option 1; the survey is not something I can run from here. In a4fbe73 the panel's guard is structural: the five sections it reads must be there and be objects, and each field it reads must be a string where it reads a string and a number where it reads a number. No ranges, no name lists, and refreshRequestLimit may be 0, null or missing. isInstanceOfConfig logs the whole validateMainConfig result with console.warn first, so a maintainer still sees the bad value, and PUT /config/main keeps every rule.

The check stays in the panel, not in the package: the field list is the panel's, not the file's, and the package should not have to know which sections a consumer reads. isMainConfig had no other caller, so it is gone from the package along with its tests.

For camera.cameraType I kept it a plain string and gave getURLPostfix a fallback to the dahua url with a console.warn. That is what the old stringToCameraType did with a name it did not know, so it is the behaviour every deployed box already had, and a wrong snapshot url costs one camera where refusing the config costs the whole panel.

instanceCheck.spec.ts covers refreshRequestLimit 0 / null / missing, a delay of 0, an unknown key, an unknown camera type, a missing section, a string where a number belongs, the unknown key next to a wrong type, and the warn. The decisions.md entry is superseded by "The panel checks the shape it reads, not the whole schema". What it does not remove: a hand-typed wrong type, backend.port: "5000" for instance, still stops the panel.

source/packages/config-schema/src/validate.test.ts
zbyju: Done in 09c484e. One table test over the eleven object paths; each sets <path>.junk = 1 on a fresh defaultConfig() and expects exactly that one error. startup is left out on purpose, it takes any key. The two single-key tests folded in. I kept the pc.first / pc.second one, because it checks the expansion into one error per key and the table does not. Checked the mutant: with every .strict() swapped for .passthrough(), twelve tests go red.

source/packages/config-schema/src/validate.test.ts
zbyju: Kept the case, moved it. isMainConfig is gone with the blocker fix, so the pair now lives in the panel's instanceCheck.spec.ts (a4fbe73). "rejects an unknown key next to a wrong type" is your first test with units.engine.ip: 10, and it goes red on a some mutant.

The second one flips to accepted: the structural guard checks cameraType as a string, and getURLPostfix falls back to the dahua url and warns. It is in the spec as "accepts a camera type it does not know", with the reason next to it.

source/apps/panel/src/utils/panel/camera.ts
zbyju: Added camera.spec.ts in a4fbe73 with both your tests, plus one for the fallback the blocker fix brought in: a name the record does not hold returns the dahua url and warns.

.github/workflows/ci.yml
zbyju: Added in e70e478, right after Build. I put a `test -d apps/backend/dist` in front of the grep, so a missing dist fails the step instead of passing it by accident. Checked against a locally built dist: clean.

source/apps/configer/src/types/main.types.test.ts
zbyju: Added in b142a67. The test moved with the other two base.json tests into services/db/main.test.ts, and it compares JSON.stringify of the parsed copy with JSON.stringify of the file. Checked: moving pc above camera in schema.ts turns it red. On the learnings wording, the line already ends "on the first PUT", so I left it as it is.

source/apps/configer/src/types/main.types.ts
zbyju: Agreed, deleted in b142a67. main.types.ts and its test file are gone, db/main.ts imports @babybox/config-schema directly, and the two base.json tests moved into services/db/main.test.ts as a describe("base.json") block together with the key-order test from the other thread. The brief asked for a thin re-export before anyone knew there would be exactly one importer, so this is where it loses.

source/apps/panel/src/defaults/config.default.ts
zbyju: Done in d2ef5e5, your version. config.default.ts is gone and the store destructures defaultConfig() once; initialised still infers as boolean. Panel tests and lint pass and the typecheck count is unchanged at 20. The "Panel defaults" row in docs/plans/config-ui.md now points at the store.

docs/decisions.md
zbyju: Done in 3264c3a. Item 2 is now one sentence about the choice and points at learnings.md, Startup. Item 3 is gone; the lockfile procedure lives in learnings.md, Tooling, and the brief already links there.

docs/decisions.md
zbyju: Done in 3264c3a. The decision entry is deleted and the two learnings are one Tooling line, close to your wording. The PR body now says the same.

docs/learnings.md
zbyju: Added in 3264c3a, under Shared package: build the package before the configer or panel tests, because both resolve it to its gitignored dist/.

docs/learnings.md
zbyju: Fixed in 3264c3a, and the same wording in the PR body.

source/packages/config-schema/src/validate.ts
zbyju: The block above validateMainConfig is gone in 3850ed2; the other thread turned that function into one line, so there was nothing left for it to explain.

I kept the L41 line. isMainConfig went away with the blocker fix, so parseMainConfig really is the one place again, and the line is what tells the next person not to cast instead.

source/packages/config-schema/src/validate.ts
zbyju: Done in 3850ed2, exactly as suggested.

source/apps/panel/src/utils/panel/camera.ts
zbyju: Fixed in a4fbe73, your wording. The file says what it saw and nothing about brands.

source/apps/backend/src/types/config.types.ts
zbyju: Done in 3850ed2, one line, as suggested. The tsconfig keeps the explanation.

source/apps/panel/src/types/panel/config.types.ts
zbyju: Deleted in 3850ed2. VoltageConfig is still used, it stays.

